PR #253, opened 2024-07-22
Chore/monorepo scripts

JoeCap08055:
Update various settings & scripts for the new monorepo structure. Tries to make as much common & top-level as possible (tsconfig, lint rules, prettier config, npm scripts, chain setup, etc)


.github/workflows/account-build.yml
```diff
@@ -9,6 +9,8 @@ on:
   push:
     branches:
       - main
+    paths:
+      - services/account
```
wilwade, line 13: Should this path also include this build file? (and same for the others here)

.gitignore
```diff
@@ -1,2 +1,135 @@
 docs/book
-.env*
+
+# Logs
```
wilwade, line 3: Should this also remove the other .gitignore? Although nested .gitignores work, but that would imply that this shouldn't be changed.

.tool-versions
```diff
@@ -0,0 +1,2 @@
+nodejs 20.12.2
```
wilwade, line 1: Remove the other ones?

.gitignore
```diff
@@ -1,2 +1,135 @@
...
+# Yarn Integrity file
+.yarn-integrity
+
+# dotenv environment variable files
+.env
+.env.docker.dev
+.env.dev
+.env.test.local
+.env.production.local
+.env.local
```
mattheworris, line 83: nit: would something like this work? When `start-gateway.sh` get ported, we also don't want to save `.env-saved`, but maybe just ignore all files that starte with `.env`?
```suggestion
.env
.env.*
```

mattheworris (approved):

.gitignore
JoeCap08055: Yeah, that's better. I just copied this wholesale from account-service.

PR #302, opened 2024-07-25
[graph] Fix extrinsic tracking

JoeCap08055:
# Description
- Update graph extrinsic monitoring queue to specify a <block-time> backoff
- Increase # of attempts to 10
- Fix bad if check for expiration # of attempts
- Add webhook endpoint specification for graph operation status


mattheworris (approved):
Looks good!
🚢 it!

PR #305, opened 2024-07-26
fix: consolidate release CI pipeline to a common workflow

JoeCap08055:
This PR refactors the release CI pipeline as follows:
- Create `release.yaml` workflow as a common release CI pipeline for all services
    - New pipeline uses a matrix approach to release one or all services based on tag prefix
    - New pipeline triggers on a "release[publish]" event instead of "push: tag"
    - Old individual service release pipelines are set to manual-only for now (will remove in a future PR)


claireolmstead (approved):
🚢 LGTM!

PR #306, opened 2024-07-26
fix: can't exclude repo checkout from matrix

JoeCap08055:
Tried to factor out the repo checkout to its own job so that it didn't have to be run multiple times, but that didn't work.


claireolmstead (approved):
Nice refactor! 🚢

PR #308, opened 2024-07-26
Refactor bull queues to a common module for shared configuration

JoeCap08055:
# Description
This PR does the following:
- Refactor multiple disparate instantiations of `BullModule.forRootAsync()` into a single, global `QueueModule`. This has the following benefits:
    - Single location for configuration of queues & default parameters
    - Proper use of NestJS dependency injection to avoid unnecessary separate instantiations of `BullModule`; cuts down on app memory profile
- Add a prefix to all Redis and BullMq cache entries to distinguish from other apps using Redis


claireolmstead (approved):
Pulling out the queue module is so nice. Nice work

PR #314, opened 2024-07-30
Formatting/lint

JoeCap08055:
- Consolidate `prettier` config to a shared package
- Consolidate `tsconfig.json` config to a shared package
- Create top-level `package.json` with a 'format' script that invokes prettier on the entire repo using the shared config
- Commit `.vscode/settings.json` containing a ref to the top-level prettier config file so that VSCode linting and formatting will work


claireolmstead (approved):
LGTM, but this seems to now be picking up some type issues with "any" and some other stuff. Let's try and resolve those warnings

claireolmstead (commented):
One more comment... since we now have a package at the top level, do we want to add an install command and format command to the top level readme?

Issue #316, opened 2024-07-30
Extract and Use Common Queue Management


wilwade:
The Account service / Reconnection Service have the best system of tracking transactions. We want to use it on:

- [x] Content Publishing Service
- [x] Graph Service

JoeCap08055:
Closed by #319

PR #319, opened 2024-07-31
Consolidate queue configs

### LINKED ISSUE #288: Queue config refactor and review
Making sure all configurations are consistent and based on queue usecase make sure the retry and all other options make sense. Duplication checks and etc.

* The configuration between producer and consumer should be consistent
* Suggestion: Using `DynamicModule` to register queues at one common module and imported in respective module

more details:
https://api.docs.bullmq.io/classes/v4.Queue.html

JoeCap08055:
For `account-service` and `content-publishing-service`:
- Reorganize imports so that we can import using '#lib/<module>' prefix instead of '../../../../lib/common/src/<module>/...'
- Refactor BullMQ connection & queue definitions to a common module for each app
    - Avoids multiple instantiation of `BullModule`
    - Eliminates possible errors in defining the same queue in multiple modules with different defaults

Closes #288 #316


services/account/libs/common/src/queues/queues.module.ts
```diff
@@ -0,0 +1,38 @@
+import { ConfigModule } from '#lib/config/config.module';
+import { ConfigService } from '#lib/config/config.service';
+import { BullModule } from '@nestjs/bullmq';
+import { DynamicModule, Global, Module } from '@nestjs/common';
+import Redis, { RedisOptions } from 'ioredis';
+import * as QueueConstants from './queue.constants';
+
+@Global()
+@Module({
```
claireolmstead, line 9: Nice!

services/content-publishing/apps/api/src/controllers/v1/asset.controller.v1.ts
```diff
@@ -22,6 +32,7 @@ export class AssetController {
     description: 'Asset files',
     type: FilesUploadDto,
   })
+  @ApiResponse({ status: '2XX', type: UploadResponseDto })
```
claireolmstead, line 35: what does this status mean?

claireolmstead (approved):
So much refactoring, it's hard to read these changes. Looks like they're all mostly names changes for imports and some rearranging, ship it! If there's any logic changes that you want me to look at, would you mind tagging me?

JoeCap08055:
> So much refactoring, it's hard to read these changes. Looks like they're all mostly names changes for imports and some rearranging, ship it! If there's any logic changes that you want me to look at, would you mind tagging me?

Sorry, I should have been more clear--no logic changes here, just refectoring of imports for shared config use and readability.

Issue #266, opened 2024-05-16
account-service: Implement endpoint to GET the `msaId` for a `publicKey`


mattheworris:
# Goal
The goal of this issue is to add an endpoint that allows callers to GET the `msaId` for a given `publicKey`. 

# Acceptance Criteria

- [x] Add new endpoint `GET/accounts/:publicKey` that returns the `msaId`

PR #325, opened 2024-08-01
add ability to get account given public key

claireolmstead:
# Problem
Currently, we cannot get an msa or any other account information given a public key. We need to add functionality to do so.

Link to GitHub Issue(s): #266 

# Solution
- Added endpoint and service to support getting an account given a public key
- Added a service to get an msaId given a public key

## Steps to Verify:

1. Spin up swagger
2. Test that endpoint return correct account.
3. See that open api displays all of the correct info.

## Screenshots (optional):
Screenshot 2024-08-01 at 3 28 19 PM

## Additional Info
While the ticket states to just return an MSA Id, I thought it would be more complete to return the associated account, which is an object that contains the msaId and handle.


services/account/apps/api/src/controllers/v1/accounts-v1.controller.ts
```diff
@@ -40,10 +40,31 @@ export class AccountsControllerV1 {
...
       this.logger.debug(`Received request to get account with msaId: ${msaId}`);
-      return await this.accountsService.getAccount(msaId);
+      return await this.accountsService.getAccountForMsaId(msaId);
+    } catch (error) {
+      this.logger.error(error);
+      throw new HttpException('Failed to find the account', HttpStatus.BAD_REQUEST);
+    }
+  }
+
+  @Get('account/:publicKey')
```
claireolmstead, line 53: Had to add a prefix so that it wouldn't get confused with the GET :msaId  endpoint. Open to changing the prefix if you guys have a better name.

services/account/apps/api/src/controllers/v1/accounts-v1.controller.ts
```diff
@@ -40,10 +40,31 @@ export class AccountsControllerV1 {
...
+   * Gets an account.
+   * @param queryParams - The query parameters for creating the account.
+   * @returns A promise that resolves to an Account object => {msaId, handle}.
+   * @throws An error if the msaId or account cannot be found.
+   */
+  async getAccountForPublicKey(@Param('publicKey') publicKey: string): Promise<AccountResponse> {
+    try {
+      this.logger.debug(`Received request to get account with publicKey: ${publicKey}`);
+      const response = await this.accountsService.getMsaIdForPublicKey(publicKey);
+      return await this.accountsService.getAccountForMsaId(response.msaId);
```
claireolmstead, line 67: Decided to return an account (account: {msaId, handle}) to make getting this data easier on the user. Happy to delete like 67 if you don't see it as being beneficial.

services/account/apps/api/src/controllers/v1/accounts-v1.controller.ts
```diff
@@ -50,6 +50,27 @@ export class AccountsControllerV1 {
...
+   * Gets an account.
+   * @param queryParams - The query parameters for creating the account.
+   * @returns A promise that resolves to an Account object => {msaId, handle}.
+   * @throws An error if the msaId or account cannot be found.
+   */
+  async getAccountForPublicKey(@Param('publicKey') publicKey: string): Promise<AccountResponse> {
+    try {
+      this.logger.debug(`Received request to get account with publicKey: ${publicKey}`);
+      const response = await this.accountsService.getMsaIdForPublicKey(publicKey);
+      return await this.accountsService.getAccount(response.msaId);
```
JoeCap08055, line 67:
```suggestion
      return this.accountsService.getAccount(response.msaId);
```

services/account/apps/api/src/services/accounts.service.ts
```diff
@@ -40,6 +40,20 @@ export class AccountsService {
...
+        this.logger.debug(`Found msaId: ${msaId} for public key: ${publicKey}`);
+        return { msaId };
+      }
+      throw new Error(`Failed to find msaId for public key: ${publicKey}`);
+    } catch (e) {
+      this.logger.error(`Error during get msaId request: ${e}`);
+      throw new Error('Failed to get msaId');
+    }
+  }
+
```
JoeCap08055, line 56: Should return `Promise<obj | null>`, and bubble up to controller, which can return a 404 if no account found for the pubkey.

services/account/apps/api/src/controllers/v1/accounts-v1.controller.ts
JoeCap08055: ☝🏻 `await` is redundant in a `return` statement

JoeCap08055 (approved):

PR #326, opened 2024-08-01
use media type as source of truth instead of type

claireolmstead:
# Problem
Need to use `mediaType` as the source of truth instead of `type` when handling assets

Link to GitHub Issue(s): 

# Solution
instead of handling an asset based on the type in the backed, the switch statement cases have been changed to handle the data based on mediaType. 

with @JoeCap08055 

## Steps to Verify:

1. What to do next
1. Any other instructions
1. Expected behavior
1. Suggestions for testing

## Screenshots (optional):


JoeCap08055:
After pairing with Aramik, I think this doesn't quite work for links, because the underlying media type could be something else. Needs further review.

JoeCap08055:
After more thought & discussion, I think the more appropriate solution for this is described in #338

claireolmstead:
close in favor of #338

PR #341, opened 2024-08-05
[bug] content-publisher needs to be its own source of truth for announcement attachments 

### LINKED ISSUE #338: [bug] content-publisher needs to be its own source of truth for announcement attachments
# Description
As a user, I should not have to keep track of the association between an uploaded asset reference ID and its media type. More specifically, I should not be able to tell content-publisher the type of an asset in an Announcment publish request (Broadcast, Reply, etc), as that opens up an attack vector to specify a media type that does not match the actual uploaded asset. content-publisher should be its own source of truth in coordinating asset type with the actual asset.

- [x] Remove `type` from `BroadcastDto`, `UpdateDto`, `ReplyDto`
   -  Removed from `AssetDto`, which is included by all of the above
- [ ] Update OpenAPI spec
   -  Updated using the `generate` scripts 
- [ ] Add `type` to cached metadata for uploaded asset

mattheworris:
# Description
As a user, I should not have to keep track of the association between an uploaded asset reference ID and its media type. More specifically, I should not be able to tell content-publisher the type of an asset in an Announcment publish request (Broadcast, Reply, etc), as that opens up an attack vector to specify a media type that does not match the actual uploaded asset. content-publisher should be its own source of truth in coordinating asset type with the actual asset.

- [x] Remove `type` from `BroadcastDto`, `UpdateDto`, `ReplyDto`
   -  Removed from `AssetDto`, which is included by all of the above
- [x] Update OpenAPI spec
   -  Updated using the `generate` scripts 
- [x] Add `type` to cached metadata for uploaded asset

Closes #338

### DIFF
--- a/services/content-publishing/README.md
+++ b/services/content-publishing/README.md
@@ -167,27 +167,16 @@ Use the provided [env.template](./env.template) file to create an initial enviro
 
   ```sh
   cp env.template .env 
-  cp env.template .env.docker.dev
   ```
 
   2. Configure the environment variable values according to your environment.
-
-### Setup
-
-Clone this repository to your desired folder:
-
-Example commands:
-
-```sh
-  git clone git@github.com:AmplicaLabs/content-publishing-service.git
-  cd content-publishing-service
-```
 
 ### Install
 
 Install NPM Dependencies:
 
 ```sh
+  cd services/content-publishing
   npm install
 ```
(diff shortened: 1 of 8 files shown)


claireolmstead (approved):
Paired on this story and tested. LGTM!!

PR #348, opened 2024-08-07
bug: Make sure blockCount is not longer than the current chain length

mattheworris:
# Problem

When running a local development frequency node, the chain length will be quite short. This results in trying to scan a blocklist that includes negative numbers, which causes the scan to fail and no previous content to be found.

# Solution
```javascript
      // Make sure blockCount is not longer than the current chain length
      if (job.data.blockCount >= startBlock) {
        job.data.blockCount = startBlock;
      }
```
## Steps to Verify:

1. Run SAT with a local frequency node and verify that the initial scan completes correctly.


claireolmstead (approved):
Debugged and created solution together. 🚢

Issue #339, opened 2024-08-05
Use dynamic registered webhooks instead of static


wilwade:
Graph services has a way to register webhooks. We want to use this and remove the static webhook environment variables.

- [ ] Account Service: Remove Static ENV Var and add per call webhook
- [ ] Content Publishing: Remove Static ENV and add per call webhook

Goal: Remove static env vars that depend on other services that can do the informing at time of call

shannonwells:
### Context
* rn you can call endpoint on content-watcher that registers a webhook for everything
* and I can submit a query w/ a webhook that tells the service to use it for the query

all gateway services should be able to do both of these.  you shouldn't have to configure the web hook in your environment.

graph service sort of has this but also has a statically configured one that should be removed and replaced with the above features.

content-publisher has none of this

additionally, ensure every service supports runtime registerable webhooks, & some endpoints should support the parametric webhook w/ a query.  the global ones are the higher priority.

emresurmeli:
This NestJS HTTP module documentation is relevant: https://docs.nestjs.com/techniques/http-module

PR #357, opened 2024-08-09
graph service should replace PROVIDER_BASE_URL with WEBHOOK_BASE_URL

mattheworris:
## Description
The gateway microservices that use webhooks previously used the environment variable `PROVIDER_BASE_URL` as the webhook target server. The `social-app-template` has adopted a new nomenclature using `WEBHOOK_BASE_URL`. All micro services should be in sync with the `social-app-template` and use the new nomenclature.

## Acceptance Criteria

- [ ] All microservices that use webhooks use `WEBHOOK_BASE_URL` to provision their webhook.

  - [ ] graph service

Follow up to remove entirely: https://github.com/ProjectLibertyLabs/gateway/issues/339


JoeCap08055:
Note, `graph-service` already does dynamic webhook registration; this `WEBHOOK_BASE_URL` is only for operating in "reconnection service mode". As such, this webhook is *actually* a _PROVIDER_ webhook, as opposed to other dynamic webhook registrations (ie, it's a fixed config). So:
1. Maybe not necessary to rename, and
2. Should consult @saraswatpuneet to understand exactly what graph service + "reconnection service required" mode is all about

saraswatpuneet (approved):
Lgtm

PR #358, opened 2024-08-09
Apply common 'prettier' formatting rules across the entire repo

JoeCap08055:
Update the individual services in the repo to reference the shared `@projectlibertylabs/prettier-config` package that lives in [packages/prettier-config](./packages/prettier-config), and apply formatting.


.prettierrc.json
```diff
@@ -1,10 +1 @@
-{
-  "arrowParens": "always",
-  "bracketSpacing": true,
-  "semi": true,
-  "trailingComma": "all",
-  "singleQuote": true,
-  "printWidth": 120,
```
wilwade, line 7: A lot of the churn in these files is moving from 180 to 120 width. Would it be worth keeping 180 to reduce the churn?

wilwade (commented):
One question about `printWidth`, but otherwise looks good to me.

wilwade (commented):
We should also have the formatting enforced on CI to make sure it actually happens before merging

.prettierrc.json
JoeCap08055: I'm fine with whatever, as long as it's consistent across the services. Though I suspect moving to 180 will cause churn in the services that had already been at 120... I'll check

JoeCap08055: As I suspected, going 120 -> 180 just churns a different set of files. Actually, 62 files vs. 61 the other way. 🤷🏻

mattheworris (approved):

PR #360, opened 2024-08-09
fix: content-watcher scanning end of chain

### LINKED ISSUE #351: content-watcher fails to find new posts when system starts
### Steps to reproduce
1. Start up SAT with a local node 
2. Make a new post.
3. Using the polkadot UI or other method, determine the block which contains the message.
4. Watch content-watcher logs and observe that it does not find anything when it scans that block.

#### Notes:
- This may also happen on testnet and does not resolve until the system is restarted.
- While repeatable, it's not clear what initial conditions are causing this problem.

### Expected behavior
content-watcher should find the messages and send an announcement to the backend.

### Actual behavior
Nothing happens, content watcher just scans blocks and never finds anything.

JoeCap08055:
This PR fixes the issue where content-watcher does not find new content at the end of the chain as it's posted.

Closes #351


mattheworris (approved):
Nice catch!
🚢 it!

Issue #354, opened 2024-05-29
[Account Service] [bug] Existing MSA w/handle login to new provider fails


JoeCap08055:
# Update Notes

The logic here is the root of the issue:
https://github.com/ProjectLibertyLabs/gateway/blob/c4372bfb5be748e7fc07511a7c3917eb43fe2584/services/account/apps/worker/src/transaction_notifier/notifier.service.ts#L176-L188

This doesn't correctly handle the case of delegation only. `handleSIWFTxnResult` catches the event for delegation, but doesn't put in all the data because it doesn't catch the not needed events of `MsaCreated` and `HandleClaimed`.

We should handle it either here, or in `handleSIWFTxnResult`:
https://github.com/ProjectLibertyLabs/gateway/blob/c4372bfb5be748e7fc07511a7c3917eb43fe2584/services/account/libs/common/src/blockchain/blockchain.service.ts#L337-L359

# Description
If an existing MSA with a handle attempts to login to a provider for which there is no current delegation, the login fails. The webhook call from the account service errors (below) and the frontend login request times out.
```
2024-05-29 09:47:58 [13:47:57.592] ERROR (198): Error handling account service webhook:
2024-05-29 09:47:58     err: {
2024-05-29 09:47:58       "type": "HttpError",
2024-05-29 09:47:58       "message": "Missing required fields",
2024-05-29 09:47:58       "stack":
2024-05-29 09:47:58           Error: Missing required fields
2024-05-29 09:47:58               at AccountServiceWebhook.accountServiceWebhook (/app/services/AccountWebhookService.ts:73:13)
2024-05-29 09:47:58               at <anonymous> (/app/controllers/WebhookController.ts:30:17)
2024-05-29 09:47:58               at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
2024-05-29 09:47:58               at WebhookController.accountServiceWebhook (/app/controllers/WebhookController.ts:29:24)
2024-05-29 09:47:58       "code": 400
2024-05-29 09:47:58     }
```

# Steps to reproduce
1. Login to SAT frontend & create a new MSA with delegation and handle
2. Login/account creation succeeds
3. In Polkadot UI (or a script), delete the delegation for the account just created
4. Logout/login again with the same account. Sign the new delegation request
5. SAT BE errors as above; login request times out in FE

PR #379, opened 2024-08-13
account service bug

claireolmstead:
# Problem
Account service was not returning all of the values when handling siwf events, causing an error when creating the webhook.

Link to GitHub Issue(s): #354 

# Solution

Added a section where we can use data that we already have to get the data that we need if it does not yet exist.

with @mattheworris 

## Steps to Verify:

1. revoke delegation
2. try and login
3. login works
4. log back in as a delegate
5. login works

### DIFF
--- a/services/account/apps/api/src/controllers/v1/accounts-v1.controller.ts
+++ b/services/account/apps/api/src/controllers/v1/accounts-v1.controller.ts
@@ -69,6 +69,7 @@ export class AccountsControllerV1 {
       const response = await this.accountsService.getMsaIdForPublicKey(publicKey);
       if (response?.msaId) {
         const account = await this.accountsService.getAccount(response.msaId);
+        this.logger.debug(account, 'Account');
         if (account) return account;
       }
       throw new HttpException('Failed to find the account', HttpStatus.NOT_FOUND);
--- a/services/account/apps/api/src/controllers/v1/accounts-v1.controller.ts
+++ b/services/account/apps/api/src/controllers/v1/accounts-v1.controller.ts
@@ -69,7 +69,6 @@ export class AccountsControllerV1 {
       const response = await this.accountsService.getMsaIdForPublicKey(publicKey);
       if (response?.msaId) {
         const account = await this.accountsService.getAccount(response.msaId);
-        this.logger.debug(account, 'Account');
         if (account) return account;
       }
       throw new HttpException('Failed to find the account', HttpStatus.NOT_FOUND);
--- a/services/account/apps/api/src/services/accounts.service.ts
+++ b/services/account/apps/api/src/services/accounts.service.ts
@@ -31,7 +31,7 @@ export class AccountsService {
           return { msaId, handle: handleResponse };
         }
         this.logger.log(`Failed to get handle for msaId: ${msaId}`);
-        return { msaId };
+        return { msaId: Number(msaId).toString() };
       }
       this.logger.log(`Invalid msaId: ${msaId}`);
       return null;
--- a/services/account/apps/api/src/services/accounts.service.ts
+++ b/services/account/apps/api/src/services/accounts.service.ts
@@ -31,7 +31,7 @@ export class AccountsService {
           return { msaId, handle: handleResponse };
         }
         this.logger.log(`Failed to get handle for msaId: ${msaId}`);
-        return { msaId: Number(msaId).toString() };
+        return { msaId };
       }
       this.logger.log(`Invalid msaId: ${msaId}`);
       return null;
--- a/services/account/apps/worker/src/transaction_notifier/notifier.service.ts
+++ b/services/account/apps/worker/src/transaction_notifier/notifier.service.ts
@@ -162,7 +162,7 @@ export class TxnNotifierService
               break;
             case TransactionType.SIWF_SIGNUP:
               {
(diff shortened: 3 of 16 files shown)


services/account/apps/api/src/controllers/v1/accounts-v1.controller.ts
```diff
@@ -69,6 +69,7 @@ export class AccountsControllerV1 {
       const response = await this.accountsService.getMsaIdForPublicKey(publicKey);
       if (response?.msaId) {
         const account = await this.accountsService.getAccount(response.msaId);
+        this.logger.debug(account, 'Account');
```
mattheworris, line 72: nit: If we keep this debug, maybe make it more traceable
```suggestion
        this.logger.debug(account, 'getAccountForPublicKey:account');
```

services/account/apps/api/src/services/accounts.service.ts
```diff
@@ -31,7 +31,7 @@ export class AccountsService {
           return { msaId, handle: handleResponse };
         }
         this.logger.log(`Failed to get handle for msaId: ${msaId}`);
-        return { msaId };
+        return { msaId: Number(msaId).toString() };
```
mattheworris, line 34: Is this needed?
```suggestion
        return { msaId };
```

services/account/apps/worker/src/transaction_notifier/notifier.service.ts
```diff
@@ -162,15 +162,15 @@ export class TxnNotifierService
...
+                const siwfTxnValues = await this.blockchainService.handleSIWFTxnResult(extrinsicEvents);
 
                 webhookResponse = createWebhookRsp(txStatus, siwfTxnValues.msaId, {
                   accountId: siwfTxnValues.address,
                   handle: siwfTxnValues.handle,
                 });
 
                 this.logger.log(
-                  `SIWF: ${siwfTxnValues.address} Signed up handle ${webhookResponse.handle} for msaId ${webhookResponse.msaId} delegated to provider ${siwfTxnValues.newProvider}.`,
+                  `SIWF: ${siwfTxnValues.address} Signed up handle ${siwfTxnValues.handle} for msaId ${webhookResponse.msaId} delegated to provider ${siwfTxnValues.newProvider}.`,
```
mattheworris, line 173: nit: suggest logging the values that are returned at the end of the function

```suggestion
                  `SIWF: ${webhookResponse.accountId} Signed up handle ${webhookResponse.handle} for msaId ${webhookResponse.msaId} delegated to provider ${webhookResponse.providerId}.`,
```

services/account/apps/worker/src/transaction_notifier/notifier.service.ts
```diff
@@ -184,7 +184,7 @@ export class TxnNotifierService
 
                 this.logger.debug(publicKeyValues.debugMsg);
                 this.logger.log(
-                  `Keys: Added the key ${webhookResponse.newPublicKey} for msaId ${webhookResponse.msaId}.`,
+                  `Keys: Added the key ${publicKeyValues.newPublicKey} for msaId ${webhookResponse.msaId}.`,
```
mattheworris, line 187:
```suggestion
                  `Keys: Added the key ${webhookResponse.newPublicKey} for msaId ${webhookResponse.msaId}.`,
```

services/account/apps/api/src/services/accounts.service.ts
claireolmstead: For some reason, it was returning a hex value when I was testing it in swagger. Not happening in the other get account so kinda confused.

services/account/swagger.json
mattheworris, line 1: I suggest we don't commit this file, as there should not be any changes to the api. I think this file is written when account-service runs and "object" is not correct, or should also describe the properties.

services/account/apps/api/src/services/accounts.service.ts
JoeCap08055: If you were getting a hex value logged there, something else is going on... this log change might just be masking a more serious issue

mattheworris (approved):
- A few non-blocking nits/comments

Nice work!

🚢 it!

services/account/swagger.json
```diff
@@ -354,7 +354,7 @@
             "type": "string"
           },
           "signature": {
-            "type": "string"
```
claireolmstead, line 357: @mattheworris what was this issue again? How should we resolve?

services/account/libs/common/src/blockchain/blockchain.service.ts
```diff
@@ -365,13 +359,28 @@ export class BlockchainService implements OnApplicationBootstrap, OnApplicationS
...
+    }
+    if (siwfTxnValues.address === '') {
+      const keyInfo = await this.getKeysByMsa(siwfTxnValues.msaId);
+      siwfTxnValues.address = keyInfo?.msa_keys[0].toString();
+    }
+    if (siwfTxnValues.newProvider === '') {
+      siwfTxnValues.newProvider = this.configService.providerId;
+    }
+
+    return siwfTxnValues;
```
JoeCap08055, line 383: Just a note that this solution works specifically because we submit our extrinsics in a batch, so whether we submit one or multiple depending on the scenario, they're all guaranteed to be in the same block.

JoeCap08055 (approved):

PR #380, opened 2024-08-13
fix: update start and stop scripts for Gateway services

### LINKED ISSUE #201: Misc Fixes from Documentation
While writing the documentation, there are some minor fixes. Here's the tracking list of them

- [x] Should be be `start-gateway.sh` or just `start.sh`
- [x] We should remove the default Provider Id from the setup script
- [x] Update "DEFAULT seed phrase needed" -> "Required for Testnet"
- [x] Remove script option for "Enter the location of the Content DB"

mattheworris:
While writing the documentation, there are some minor fixes. Here's the tracking list of them

- [x] Should be be `start-gateway.sh` or just `start.sh`
- [x] We should remove the default Provider Id from the setup script
- [x] Update "DEFAULT seed phrase needed" -> "Required for Testnet"
- [x] Remove script option for "Enter the location of the Content DB"

Closes #201


start.sh
```diff
@@ -22,7 +22,8 @@ if [ -f .env-saved ]; then
     echo -e "Found saved environment from a previous run:\n"
     cat .env-saved
     echo
-    read -p  "Do you want to re-use the saved paramters? [y/N]: " REUSE_SAVED
+    read -p  "Do you want to re-use the saved paramters? [Y/n]: " REUSE_SAVED
```
mattheworris, line 25: Changed to match SAT

start.sh
```diff
@@ -116,11 +116,12 @@ EOI
     cat << EOI
 
 ┏━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┓
-┃ IPFS settings                                                                               ┃
+┃ Do you want to change the IPFS settings [y/N]:                                              ┃
```
mattheworris, line 119: Changed to match SAT

start.sh
```diff
@@ -148,14 +150,6 @@ then
     # Start specific services in detached mode
     echo -e "\nStarting local frequency services..."
     docker compose up -d frequency
-
-    # Wait for 15 seconds
```
mattheworris, line 152: I don't think this is necessary for gateway now.

start.sh
```diff
@@ -116,11 +116,12 @@ EOI
     cat << EOI
 
 ┏━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┓
-┃ IPFS settings                                                                               ┃
+┃ Do you want to change the IPFS settings [y/N]:                                              ┃
+┃ [When using Testnet, you must change the IPFS Provider information]                         ┃
```
wilwade, line 120: Noting a connection to the update on the SAT side of this: https://github.com/ProjectLibertyLabs/social-app-template/pull/146/files#r1716891837

claireolmstead:
Only question here... if we're pointing out that the seed phrase is required for testnet, we should probably point out that the provider id and some of the IPFS things are also required. 

Alternate solution: Is there a way we can change the default value based on if they select local or testnet?

mattheworris:
> Only question here... if we're pointing out that the seed phrase is required for testnet, we should probably point out that the provider id and some of the IPFS things are also required.
> 
> Additional solution: Change the IPFS default values based on if they select local or testnet?

`PROVIDER_ID` default is also "INPUT REQUIRED"

The IPFS defaults have been updated so that they should work with local and testnet. 
(tracking an issue on SAT where IPFS links are not using the IPFS UA gateway to resolve, but I think all the gateway services are happy with these defaults)

JoeCap08055 (approved):

PR #386, opened 2024-08-14
Fix OpenAPI document generation

JoeCap08055:
Update the embedded environment used when running the app to generate the OpenAPI spec document


mattheworris (approved):
Looks good!
🚢 it!

PR #393, opened 2024-08-16
fix: resume queue in content-publishing when Capacity is available

### LINKED ISSUE #392: [bug] [content-publishing] Queue does not resume on Capacity refill
# Description
The publishing queue is paused when Capacity is exhausted, and is supposed to be resumed once Capacity becomes available (either at the next epoch, or if more Capacity is staked).

# Expected Results
- Queue is resumed once Capacity becomes available

# Actual Results
- Message is logged that Capacity is available, but queue is not resumed

JoeCap08055:
# Description
Resume queue when Capacity is available (was logging event, but not resuming queue)

# Acceptance Criteria
- [x] Publishing queue is resumed when Capacity becomes available

Closes #392

### DIFF
--- a/services/content-publishing/apps/worker/src/publisher/publishing.service.ts
+++ b/services/content-publishing/apps/worker/src/publisher/publishing.service.ts
@@ -112,6 +112,7 @@ export class PublishingService extends BaseConsumer implements OnApplicationBoot
     if (await this.publishQueue.isPaused()) {
       this.logger.debug('Received capacity.available event');
     }
+    await this.publishQueue.resume();
     try {
       this.schedulerRegistry.deleteTimeout(CAPACITY_EPOCH_TIMEOUT_NAME);
     } catch (err) {


saraswatpuneet (approved):
👍🏽

mattheworris (approved):

services/content-publishing/apps/worker/src/publisher/publishing.service.ts
```diff
@@ -112,6 +112,7 @@ export class PublishingService extends BaseConsumer implements OnApplicationBoot
     if (await this.publishQueue.isPaused()) {
       this.logger.debug('Received capacity.available event');
     }
+    await this.publishQueue.resume();
```
saraswatpuneet, line 115: Move it inside try ?

JoeCap08055: I don't think it's necessary; the try is there to catch & ignore if the timeout is non-existent; other errors should bubble up.

JoeCap08055: I believe `resume()` is a no-op if the queue is not paused

PR #420, opened 2024-08-22
Separate nonce service Redis connection and track Redis connection status correctly

### LINKED ISSUE #417: [bug] Nonce service should be a singleton across Gateway
# Description
Each Gateway service makes use of a `NonceService` class that utilizes Redis to keep track of submitted extrinsics & nonce usage, so that multiple extrinsic submissions in a short period of time don't fall afoul of either duplicate nonces or holes in the nonce.

Because each service uses the same keypair/seed phrase, the Redis key used for the NonceService should be common across the services; currently the service-specific key prefix applies to the nonce service keys as well, which makes it more likely that the services could encounter nonce problems.

The key used should incorporate the public key of the provisioned keypair, as nonces are tied to a public key, not an MSA.

JoeCap08055:
This PR does 2 things:
1. Separate out the Redis connection for `NonceService` in all of the Gateway services, so that the key prefix is global & includes the Provider's public key address.
2. Refactor the Redis connection status monitoring into a service class & refactor the main app function, especially in the worker apps, to properly initialize & await things so that emitted events work properly.

Closes #417


services/account/apps/api/src/api.module.ts
```diff
@@ -43,28 +42,17 @@ import { ApiService, AccountsService, HandlesService, DelegationService, KeysSer
...
+      useFactory: (configService: ConfigService) => [
+        {
+          url: configService.redisUrl.toString(),
+          keyPrefix: configService.cacheKeyPrefix,
+          maxRetriesPerRequest: null,
+        },
+      ],
+      inject: [ConfigService],
+    }),
+    QueueModule,
```
JoeCap08055, line 55: 1:1 swap, `RedisModule` is now initialized inside `CacheModule`. This pattern across all Gateway apps & services.

services/account/apps/api/src/main.ts
JoeCap08055, line 1: `bootstrap` method of all apps refactored as follows:
* Remove all Redis connection monitoring; that now lives in `CacheMonitorService`
* Make sure we always call `app.listen()`, even for Worker apps with no endpoints. This required creating a `KeepAliveStrategy` class to be able to instantiate the app as a NestMicroservice without actually binding to a port.
* Add a shutdown timer to make sure the process actually terminates when we ask it (@nestjs/bullmq is not always well-behaved when shutting down)

This pattern replicated across all Gateway main apps (api + worker)

services/account/apps/api/src/services/api.service.ts
JoeCap08055, line 1: Dummy class did nothing; removed

services/account/apps/worker/src/transaction_notifier/notifier.service.ts
```diff
@@ -1,5 +1,5 @@
 /* eslint-disable no-await-in-loop */
-import { InjectRedis } from '@liaoliaots/nestjs-redis';
+import { InjectRedis } from '@songkeys/nestjs-redis';
```
JoeCap08055, line 2: Upgraded `account-service` to NestJS v10.x, which also required changing the redis package

services/account/libs/common/src/cache/cache-monitor.service.ts
JoeCap08055, line 1: Monitoring logic for Redis connections now lives here; instead of emitting `redis.*` events for `main.bootstrap` to catch, we handle everything here and simply emit a `shutdown` event if appropriate.

services/account/libs/common/src/cache/cache.module.ts
JoeCap08055, line 1: This module encapsulates initializing `RedisModule` and wiring up the event listeners to react to connection status.

services/account/libs/common/src/config/config.service.ts
JoeCap08055, line 1: All `ConfigService` classes modified to have the Provider's public key address, which necessitates awaiting the readiness of the crypto subsystem on startup.

services/account/libs/common/src/utils/keepalive-strategy.ts
JoeCap08055, line 1: This class allows us to have our `worker` apps initialize as NestMicroservice, so that Schedulers and EventEmitters work correctly.

services/account/apps/api/src/api.module.ts
JoeCap08055, line 1: Note on reviewing--changes are mirrored across all Gateway services; it's a big PR, but the changes are duplicated, so you only really need to dive deep into a single service.

services/account/apps/worker/src/main.ts
```diff
@@ -13,85 +13,49 @@ BigInt.prototype['toJSON'] = function () {
...
-    logger: process.env.DEBUG ? ['error', 'warn', 'log', 'verbose', 'debug'] : ['error', 'warn', 'log'],
-  });
-  logger.log('Nest ApplicationContext created successfully.');
-
   process.on('uncaughtException', (error) => {
     console.error('****** UNCAUGHT EXCEPTION ******', error);
     process.exit(1);
   });
 
+  const app = await NestFactory.createMicroservice(WorkerModule, {
```
mattheworris, line 32: Just curious, is this new for v10?

services/content-publishing/apps/api/src/main.ts
```diff
@@ -21,8 +33,10 @@ async function bootstrap() {
 
   // Get event emitter & register a shutdown listener
   const eventEmitter = app.get<EventEmitter2>(EventEmitter2);
+  eventEmitter.setMaxListeners(20);
```
mattheworris, line 36: Do we also need to increment the max listeners in account-service?

JoeCap08055: Probably not; `content-publishing` was logging warning & errors about leaking resources due to exceeding max listeners, it think mainly due to the number of queues and things. `account-service` isn't issuing that warning, so probably fine.

services/account/libs/common/src/cache/cache-monitor.service.ts
```diff
@@ -0,0 +1,68 @@
...
+
+  public registerRedisClient(client: Redis, namespace: ClientNamespace = 'default') {
+    client.on('error', (error) => this.handleRedisError(error, namespace));
+    client.on('close', () => this.handleRedisClose(namespace));
+    client.on('ready', () => this.handleRedisReady(namespace));
+  }
+
+  // eslint-disable-next-line class-methods-use-this
+  public handleRedisError(err: unknown, namespace: ClientNamespace) {
+    // this.logger.error(`REDIS ERROR [${namespace.toString()}]: `, err);
```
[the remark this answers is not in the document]
JoeCap08055: It was super noisy and not very helpful.

services/account/apps/worker/src/main.ts
JoeCap08055: No, it always existed. Using it was recommended as a solution for how to spin up a NestJS app that doesn't actually listen on any ports or connections, but still runs continuously as a service, while also having it properly initialize the NestJS-based frameworks surrounding schedulers and event emitters.

services/graph/apps/worker/src/worker.module.ts
```diff
@@ -34,18 +26,34 @@ import { QueueModule } from '#lib/queues/queue.module';
       // set this to `true` if you want to emit the removeListener event
       removeListener: false,
       // the maximum amount of listeners that can be assigned to an event
-      maxListeners: 10,
+      maxListeners: 20,
```
mattheworris, line 29: Ah, here's the `maxListeners` update for graph. Is this tuned per service? I see it's either 10 or 20.

mattheworris (approved):
- Read through changes
- A few non-blocking comments.

Nice work!
🚢 it!

services/graph/apps/worker/src/worker.module.ts
JoeCap08055: I probably tuned it based on whether I saw an error/warning or not.

JoeCap08055: (I believe the default is 10)

PR #444, opened 2024-08-26
feat: at startup, check to make sure provider seed phrase matches configured provider ID

### LINKED ISSUE #419: Services should validate provider seed phrase at startup
# Description
As a user/someone who is deploying Gateway, I want to be alerted if there are errors in my configuration. Specifically, I want to know if I've mis-configured my seed phrase or Provider ID.

## Detail
When a service starts up, it should perform some validation on the `PROVIDER_SEED_PHRASE`. Possible options:
1. Validate that it's a valid seed phrase
2. Generate the public key and query the chain for the Provider MSA
3.
    a. Validate the returned MSA against the configured `PROVIDER_ID` ---OR---
    b. Eliminate `PROVIDER_ID` as a configuration value, and populate it at startup from the chain

## Notes
Since we are expecting to add a "read-only" deployment mode to Gateway, which would eliminate the requirement for a seed phrase in that mode, it might make more sense to pursue option `3a` rather than `3b`. Discuss in comments.

JoeCap08055:
- Adds logic to startup in `account`, `content-publishing`, `content-watcher` to query the chain for the MSA ID associated with the public key derived from the configured seed phrase; throw an error if it does not match the configured Provider ID
- Syncs initialization code for `BlockchainService` in `graph-service` with the rest of the services' implementations (missed in previous PR)

Closes #419

### DIFF
--- a/services/graph/libs/common/src/blockchain/blockchain.service.ts
+++ b/services/graph/libs/common/src/blockchain/blockchain.service.ts
@@ -37,6 +37,13 @@ export class BlockchainService implements OnApplicationBootstrap, BeforeApplicat
 
   private logger: Logger;
 
+  private readyResolve: (boolean) => void;
+  private readyReject: (reason: any) => void;
+  private isReadyPromise = new Promise<boolean>((resolve, reject) => {
+    this.readyResolve = resolve;
+    this.readyReject = reject;
+  });
+
   public async onApplicationBootstrap() {
     const providerUrl = this.configService.frequencyUrl!;
     let provider: any;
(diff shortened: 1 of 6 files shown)


services/graph/libs/common/src/blockchain/blockchain.service.ts
```diff
@@ -254,4 +261,19 @@ export class BlockchainService implements OnApplicationBootstrap, BeforeApplicat
...
+
+  public async publicKeyToMsaId(publicKey: string): Promise<string | null> {
+    const handleResponse = await this.query('msa', 'publicKeyToMsaId', publicKey);
+    if (handleResponse.isSome) return handleResponse.unwrap().toString();
+    return null;
+  }
+
+  public async validateProviderSeedPhrase() {
+    const { providerPublicKeyAddress, providerId } = this.configService;
+    const resolvedProviderId = await this.publicKeyToMsaId(providerPublicKeyAddress || '');
```
wilwade, line 273: Does this not error out with the `''` version?

wilwade (approved):
One question and a test would be nice, but it is a small piece of code that is quite clear.

mattheworris (approved):
🚢 it!

wilwade (approved):
Nice improvements

PR #446, opened 2024-08-27
Enable read-only mode for account & graph services

### LINKED ISSUE #418: Read Only Modes
Several of the Gateway services are useful without needing to transact on the chain. These services currently require a key to start.

This change is to make them able to start in a read only mode where a key is not required and the endpoints that would require the key are disabled in some way.

Services:
- [ ] Account Service
- [ ] Graph Service

## Options
- Read only mode enabling:
    - Just start in read only mode if the key and provider id is not provided
    - Require enabling read only mode via an env var which then makes the key and provider id not required
- Endpoints requiring writing to the chain when in read only mode
    - Be completely disabled and "not exist"
    - Exist, but return 400 or 404 and log that the device is in read only mode.

Startup logs should note the mode the service is in either way.

JoeCap08055:
# Description
- Allow `PROVIDER_ACCOUNT_SEED_PHRASE` to be missing/empty for `api` apps in `acount-service` and `graph-service`. 
- Add getter `isReadOnly` to ConfigService; returns `true` If `PROVIDER_ACCOUNT_SEED_PHRASE` is missing/empty in an api service
- Add a `ReadOnlyGuard` guard to API routes that result in a chain modification (ie, creating some task that submits an extrinsic). If `ConfigService.isReadOnly === true`, then such endpoints return `HTTP 403 (FORBIDDEN)`
- Add/update tests

Closes #418


services/account/apps/api/src/guards/read-only.guard.ts
```diff
@@ -0,0 +1,25 @@
...
+  // eslint-disable-next-line no-useless-constructor, no-empty-function
+  constructor(private configService: ConfigService) {}
+
+  canActivate(context: ExecutionContext): boolean {
+    const { isReadOnly } = this.configService;
+    if (isReadOnly) {
+      // If the system is in read-only mode, disable certain routes
+      const request = context.switchToHttp().getRequest();
+      const { method } = request;
+      const readOnlyMethods = ['POST', 'PUT', 'DELETE']; // Methods that should be restricted
```
mattheworris, line 15: nit: this name implies the opposite to me, that these are the allowed methods for read-only
```suggestion
      const restrictedMethods = ['POST', 'PUT', 'DELETE']; // Methods that should be restricted
```

services/account/apps/api/src/guards/read-only.guard.ts
```diff
@@ -0,0 +1,25 @@
+import { ConfigService } from '#lib/config';
+import { Injectable, CanActivate, ExecutionContext } from '@nestjs/common';
+
+@Injectable()
+export class ReadOnlyGuard implements CanActivate {
```
mattheworris, line 5: Nice use of guards.

mattheworris (approved):
- Read through changes
- Nice feature add
- non-blocking comment #NamingIsHard
🚢 it!

Issue #413, opened 2024-08-19
`changeHandle`


JoeCap08055:
# Description
As a Provider/consumer of the Gateway API, I want to be able to submit a signed `changeHandle` request on behalf of a user

Should implement:
- [x] `GET /account/handles/changeHandle/{newHandle}?expirationSeconds={expirationSeconds}` (returns a properly encoded `ClaimHandlePayload` that can be signed)
    - [ ] encoded payload should be a hex-encoded, byte-wrapped byte array
    - [ ] default expiration time should be 10 minutes if not provided in request (to match SIWF)
- [x] `PUT /account/handles/changeHandle` accepts a signed `ClaimHandlePayload` for submission via a `changeHandle` extrinsic

PR #458, opened 2024-08-29
413 changehandle

claireolmstead:
# Problem

We do not have an endpoint to return an encoded payload for signing a change handle call.

Link to GitHub Issue(s): #413 

# Solution

- Create GET endpoint to return an encoded payload.
- Remove encoding from the POST change handle.

with @JoeCap08055 @enddynayn 

## Steps to Verify:

1. Get encoded extrinsic
2. Post change handle

### DIFF
--- a/services/account/apps/api/src/controllers/v1/handles-v1.controller.ts
+++ b/services/account/apps/api/src/controllers/v1/handles-v1.controller.ts
@@ -8,20 +8,23 @@ import {
   Param,
   HttpException,
   Body,
-  UseGuards,
 } from '@nestjs/common';
 import { ApiBody, ApiOkResponse, ApiOperation, ApiTags } from '@nestjs/swagger';
 import { TransactionType } from '#lib/types/enums';
 import { HandlesService } from '#api/services/handles.service';
 import { EnqueueService } from '#lib/services/enqueue-request.service';
+import {
+  ChangeHandlePayloadRequest,
+  ChangeHandleRequest,
+  CreateHandleRequest,
+} from '#lib/types/dtos/handles.request.dto';
 import { ChangeHandleRequest, CreateHandleRequest, HandleRequestDto } from '#lib/types/dtos/handles.request.dto';
 import { TransactionResponse } from '#lib/types/dtos/transaction.response.dto';
+import { u8aToHex, u8aWrapBytes } from '@polkadot/util';
 import { HandleResponseDto } from '#lib/types/dtos/accounts.response.dto';
-import { ReadOnlyGuard } from '#api/guards/read-only.guard';
 
 @Controller('v1/handles')
 @ApiTags('v1/handles')
-@UseGuards(ReadOnlyGuard) // Apply guard at the controller level
 export class HandlesControllerV1 {
   private readonly logger: Logger;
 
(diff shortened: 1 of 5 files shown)


services/account/apps/api/src/metadata.ts
```diff
@@ -110,19 +114,23 @@ export default async () => {
...
         ],
         [
-          import('../../../libs/common/src/types/dtos/transaction.response.dto'),
-          { TransactionResponse: { referenceId: { required: true, type: () => String } } },
-        ],
-        [
-          import('../../../libs/common/src/types/dtos/wallet.login.response.dto'),
+          import('../../../libs/common/src/types/dtos/delegation.response.dto'),
           {
-            WalletLoginResponse: {
```
claireolmstead, line 119: I didn't change any of this code. was the metadata just out of date?

services/account/apps/api/src/controllers/v1/handles-v1.controller.ts
```diff
@@ -69,6 +75,32 @@ export class HandlesControllerV1 {
     }
   }
 
+  @Get('change/:newHandle')
+  @HttpCode(HttpStatus.OK)
+  @ApiOperation({ summary: 'Get a properly encoded ClaimHandlePayload that can be signed.' })
+  @ApiOkResponse({ description: 'ClaimHandlePayload request enqueued' })
```
JoeCap08055, line 81:
```suggestion
  @ApiOkResponse({ description: 'Returned an encoded ClaimHandlePayload for signing' })
```

services/account/apps/api/src/services/handles.service.ts
```diff
@@ -17,4 +19,13 @@ export class HandlesService {
     }
     throw new Error('Invalid msaId.');
   }
+
+  async getExpiration(): Promise<number> {
+    const lastFinalizedBlockNumber = await this.blockchainService.getLatestFinalizedBlockNumber();
+    return lastFinalizedBlockNumber + 10;
+  }
+
+  async encodePayload(payload: HandleRequest['payload']): Promise<CommonPrimitivesHandlesClaimHandlePayload> {
```
JoeCap08055, line 28:
```suggestion
 encodePayload(payload: HandleRequest['payload']): CommonPrimitivesHandlesClaimHandlePayload {
```

services/account/apps/api/src/controllers/v1/handles-v1.controller.ts
```diff
@@ -69,6 +75,32 @@ export class HandlesControllerV1 {
...
+   * Using the provided query parameters, creates a new payload that can be signed to change handle.
+   * @param queryParams - The query parameters for changing the handle.
+   * @returns Payload is included for convenience. Encoded payload to be used when signing the transaction.
+   * @throws An error if the change handle payload creation fails.
+   */
+  async getChangeHandlePayload(@Param('newHandle') newHandle: string): Promise<ChangeHandlePayloadRequest> {
+    try {
+      const expiration = await this.handlesService.getExpiration();
+      const payload = { baseHandle: newHandle, expiration };
+      const encodedPayload = await this.handlesService.encodePayload(payload);
```
JoeCap08055, line 92: So, this gives us a SCALE-encoded payload. But we need to sign raw bytes. So what we actually need to return is a Byte-wrapped array of bytes, encoded as a hex string. Like so:
```suggestion
      const encodedPayload = u8aToHex(u8aWrapBytes(this.handlesService.encodePayload(payload).toU8a()));
```

Also, `encodePayload` shouldn't be `async`, so we don't need `await`.

services/account/apps/api/src/services/handles.service.ts
```diff
@@ -17,4 +19,13 @@ export class HandlesService {
     }
     throw new Error('Invalid msaId.');
   }
+
+  async getExpiration(): Promise<number> {
+    const lastFinalizedBlockNumber = await this.blockchainService.getLatestFinalizedBlockNumber();
+    return lastFinalizedBlockNumber + 10;
```
JoeCap08055, line 25: FYI the standard expiration we adopted in SIWF for claiming a handle is 10 minutes (= 600 / SECONDS_PER_BLOCK) blocks.

Though I wonder if (time in seconds) should be an (optional) input parameter to the `GET` endpoint; if not provided we could assume 10 minutes. I think we should add that; it's easy enough. If there are no objections @mattheworris @claireolmstead , I'll add it to the issue AC.

services/account/libs/common/src/types/dtos/handles.request.dto.ts
```diff
@@ -28,6 +30,16 @@ export class HandleRequest {
   proof: HexString;
 }
 
+export class ChangeHandlePayloadRequest {
+  @ApiProperty()
+  @IsNotEmpty()
```
JoeCap08055, line 35: FYI I don't think these `class-validator` decorators have any effect on return types; only on input types.

services/account/swagger.json
JoeCap08055, line 1: Don't forget to re-generate this file again after making more changes to the controller.

JoeCap08055 (changes requested):

JoeCap08055 (approved):

services/account/apps/api/src/services/handles.service.ts
```diff
@@ -17,4 +17,13 @@ export class HandlesService {
...
```
mattheworris, line 23: 10 blocks may not be long enough to enable signing of this transaction

100@ 6 seconds or 50 @ 12 seconds should probably be a default constant from the config service
```suggestion
    return lastFinalizedBlockNumber + 100;
```
or if we add `@Query('expirationTime') expirationTime?: number,` to the endpoint.
```suggestion
    return lastFinalizedBlockNumber + expirationTime/SECS_PER_BLOCK; //Pseudocode
```

mattheworris (changes requested):
- Read through changes
- Requesting updates to block expiration

mattheworris (approved):

Issue #412, opened 2024-08-19
`retireMsa`


JoeCap08055:
# Description
As a consumer of the Gateway API, I want to be able to pass along a signed transaction to retire an MSA on behalf of a user.

Should implement:
`GET /account/accounts/retireMsa` (returns a `retireMsa` encoded extrinsic that can be signed)
`POST /account/accounts/retireMsa` submits a signed extrinsic for execution

PR #472, opened 2024-09-03
412 retiremsa

claireolmstead:
# Problem

As a user, we want to be able to retire an MSA ID.

Link to GitHub Issue(s): #412 

# Solution

- Create an endpoint to get an easy to use retire msa payload.
- Create an endpoint to post a retire msa payload that adds the transaction to the queue and executes it.

with @wilwade @mattheworris @JoeCap08055 @enddynayn 

## Steps to Verify:

1. Create an Msa for bob
2. Test through the `apps/account-api/test/setup/test-retire-msa.ts` file, making sure it successfully executes
3. Test through the swagger endpoints. Note: you'll have to generate a correct signature.


services/content-publishing/swagger.json
JoeCap08055, line 1: This file should be excluded from this PR

services/account/libs/common/src/types/dtos/accounts.request.dto.ts
```diff
@@ -0,0 +1,27 @@
...
+import { SignerPayloadJSON } from '@polkadot/types/types';
+
+export class RetireMsaRequestDto {
+  @ApiProperty({ type: Object })
+  @IsNotEmpty()
+  unsignedPayload: SignerPayloadJSON;
+
+  @ApiProperty({ type: String })
+  @IsNotEmpty()
+  encodedPayload: ExtrinsicPayload;
```
JoeCap08055, line 14: In general, I don't think we should use types from the `@polkadot/*` libraries directly in our REST APIs. The reason being twofold:
1. The types are subject to change outside of our control, thus changing our API potentially unintentionally
2. The types are outside our control; we don't have the ability to add `@ApiProperty()` decorators to them & therefore our generated OpenAPI spec is opaque when it comes to these fields
3. These types are SCALE-encoded types; although there is some work going on under the hood to convert to standard JSON non-SCALE representation when we respond, in general it's not a good practice, as we're trying to isolate the client from having exposure to SCALE concerns.

I'd suggest creating our own types for any request/response objects, and if we want to marshal to/from the `@polkadot/*` types, we should create corresponding `to...` and `from...` utility methods.

apps/account-api/src/controllers/v1/accounts-v1.controller.ts
```diff
@@ -57,7 +60,7 @@ export class AccountsControllerV1 {
     }
   }
 
-  @Get('account/:publicKey')
+  @Get('account/:accountId')
```
claireolmstead, line 63: name change for consistency.

apps/account-api/src/services/accounts.service.ts
```diff
@@ -41,7 +47,7 @@ export class AccountsService {
     }
   }
 
-  async getMsaIdForPublicKey(publicKey: string): Promise<MsaIdResponse | null> {
+  async getMsaIdForAccountId(publicKey: string): Promise<MsaIdResponse | null> {
```
claireolmstead, line 50: name change for consistency

apps/account-api/test/accounts.controller.e2e-spec.ts
```diff
@@ -95,4 +97,38 @@ describe('Account Controller', () => {
       .expect((res) => res.body.msaId === validMsaId)
       .expect((res) => res.body.handle.base_handle === handle);
   });
+
+  it('(GET) /v1/accounts/retireMsa/:accountId get payload for retireMsa, given a valid accountId', async () => {
```
claireolmstead, line 101: what should we do about these test?

apps/account-api/src/services/accounts.service.ts
```diff
@@ -108,4 +114,21 @@ export class AccountsService {
     }
     throw new Error('Invalid Sign In With Frequency Request');
   }
+
+  getRetireMsaPayload(accountId: string): Promise<{ signerPayload: SignerPayloadRaw; encodedPayload: string }> {
+    return this.blockchainService.createRetireMsaPayload(accountId);
+  }
```
JoeCap08055, line 120: Question for discussion:

Should we proactively check some pre-conditions here before generating the extrinsic? Specifically:
- Check that `accountId` maps to a valid MSA
- Check that the MSA is eligible to be retired (ie, no active delegations)

These checks are done by the extrinsic itself, but I'm wondering if it's beneficial to do them here as well to avoid submitting an extrinsic that could ultimately fail? Granted, it's free, and we have a signature replay registry in the node to mitigate a replay attack, I believe... but I'm still thinking it's not a bad idea here?

Thoughts?

apps/account-api/test/accounts.controller.e2e-spec.ts
JoeCap08055: What's the blocker with getting them to work correctly?

apps/account-api/test/accounts.controller.e2e-spec.ts
```diff
@@ -95,4 +97,38 @@ describe('Account Controller', () => {
...
+      .expect(200)
+      .expect((res) => res.body.unsignedPayload === '')
+      .expect((res) => res.body.encodedPayload === '');
+  });
+
+  // if (
+  //   ('(GET) /v1/accounts/retireMsa/:accountId get payload for retireMsa, given an invalid accountId', async () => {})
+  // );
+
+  it.only('(POST) /v1/accounts/retireMsa post retireMsa', async () => {
```
JoeCap08055, line 114: Remove `.only`

JoeCap08055: If there are other tests in this file that are broken, we should either (a) fix them, or (b) mark the broken ones with `.skip`, but let's not commit `.only`

apps/account-api/test/accounts.controller.e2e-spec.ts
```diff
@@ -95,4 +97,38 @@ describe('Account Controller', () => {
...
+
+    const uint8Signature = keypair.sign(JSON.stringify(responseData.unsignedPayload), { withType: false });
+    const signature = u8aToHex(uint8Signature);
+
+    const retireMsaRequest = {
+      unsignedPayload: responseData.unsignedPayload,
+      encodedPayload: responseData.encodedPayload,
+      signature,
+      accountId,
+    };
```
JoeCap08055, line 131: The structure of the response data here doesn't seem to match the endpoint:
```
{
   signerPayload: SignerPayloadRaw,
   encodedPayload: string,
}
```

apps/account-api/test/setup/test-retire-msa.ts
JoeCap08055, line 1: Shouldn't this just be incorporated into the e2e tests, instead of a separate script?
(realizing this was probably useful during heavy debugging, but now that we've got it working it's just one more separate thing to maintain...)

apps/account-worker/src/transaction_publisher/publisher.service.ts
```diff
@@ -196,6 +204,34 @@ export class TransactionPublisherService extends BaseConsumer implements OnAppli
     }
   }
 
+  async processProxyTxn(
+    ext: any,
```
JoeCap08055, line 208: Is it possible to use some more specific type information here instead of `any`?

apps/account-worker/src/transaction_publisher/publisher.service.ts
```diff
@@ -196,6 +204,34 @@ export class TransactionPublisherService extends BaseConsumer implements OnAppli
...
+    accountId: string,
+    signer: Signer,
+    signerPayload: SignerPayloadRaw,
+  ): Promise<HexString> {
```
JoeCap08055, line 212: This should return both the signed & submitted extrinsic itself, as well as the hash, like `[SubmittableExtrinsic<'promise'>, HexString]` (see `processSingleTxn` and `processBatchTxn`). This will allow us to track the extrinsic to completion using its mortality era.

docker-compose.yaml
```diff
@@ -56,15 +56,15 @@ services:
...
     # We need to specify the platform because it's the only image
     # built by Frequency at the moment, and auto-pull won't work otherwise
     platform: linux/amd64
     # Uncomment SEALING_MODE and SEALING_INTERVAL if you want to use interval sealing.
     # Other options you may want to add depending on your test scenario.
     environment:
-      - SEALING_MODE=interval
-      - SEALING_INTERVAL=12
+      - SEALING_MODE=instant
+      # - SEALING_INTERVAL=12
```
JoeCap08055, line 67: We should avoid using instant seal in our test setups. But we can set the `SEALING_INTERVAL` to a lower number for test speed.

libs/account-lib/src/blockchain/blockchain.service.ts
```diff
@@ -511,4 +510,50 @@ export class BlockchainService implements OnApplicationBootstrap, OnApplicationS
...
+            signRaw = raw;
+            // Interrupt the signing process to get the raw payload, as encoded by polkadot-js
+            throw new Error('Stop here');
+          },
+        },
+      });
+    } catch (_e) {
+      return signRaw;
+    }
+  }
```
JoeCap08055, line 535: Thanks a lot, Parity, for making it necessary to do this kind of hokey stuff to get a signing payload 😬 

I think this function can be tweaked as to how the return value is collected so that we don't need to disable lint rules. Also, it can be a static method since it doesn't use any of the class properties (otherwise we could ignore the lint rule for that). Something like:
```suggestion
  public static async getRawPayloadForSigning(
    tx: SubmittableExtrinsic<'promise', ISubmittableResult>,
    signerAddress: string,
  ): Promise<SignerPayloadRaw> {
    const fakeError = '*** Interrupt signing for payload collection ***';
    let signRaw: SignerPayloadRaw;
    try {
      await tx.signAsync(signerAddress, {
        signer: {
          signRaw: (raw) => {
            console.log('signRaw called with [raw]:', raw);
            signRaw = raw;
            // Interrupt the signing process to get the raw payload, as encoded by polkadot-js
            throw new Error(fakeError);
          },
        },
      });
    } catch (e: any) {
      // If we encountered an actual error, re-throw it; otherwise
      // ignore the fake error we threw above
      if (e?.message !== fakeError) {
        throw e;
      }
    }

    return signRaw;
  }
```

libs/account-lib/src/types/dtos/accounts.request.dto.ts
```diff
@@ -0,0 +1,26 @@
...
+  @IsNotEmpty()
+  signerPayload: SignerPayloadRaw;
+
+  @ApiProperty({ type: String })
+  @IsNotEmpty()
+  encodedPayload: string;
+
+  @ApiProperty({ type: Object })
+  @IsNotEmpty()
+  signer: Signer;
```
JoeCap08055, line 17: Suggestion: for `signerPayload` and `signer`, let's create a concrete `class` that implements each interface; this way we can actually use that type in the `@ApiProperty({ type: xxx })` decorator, and the OpenAPI spec will contain the correct structure for the payload, rather than just an opaque `Object`.

I think this is necessary for a useful public API spec.

libs/graph-lib/src/services/nonce.service.ts
JoeCap08055, line 1: Not sure why this file changed as part of this PR...

openapi-specs/account.openapi.json
JoeCap08055, line 1: I think this file needs to be re-generated; it doesn't match the DTO in the code.

apps/account-api/src/services/accounts.service.ts
enddynayn: That’s a great idea! I think we could also update the name of the parameter, maybe to something like "account."

openapi-specs/account.openapi.json
JoeCap08055: Make sure to re-generate the metadata as well.

JoeCap08055 (changes requested):
Great work--I know it was quite a slog getting everything working!
Just a few tweaks needed; see my other comments.

apps/account-api/test/accounts.controller.e2e-spec.ts
```diff
@@ -95,4 +97,38 @@ describe('Account Controller', () => {
...
+  it('(GET) /v1/accounts/retireMsa/:accountId get payload for retireMsa, given a valid accountId', async () => {
+    const accountId = users[0].keypair.address;
+    await request(app.getHttpServer())
+      .get(`/v1/accounts/retireMsa/${accountId}`)
+      .expect(200)
+      .expect((res) => res.body.unsignedPayload === '')
+      .expect((res) => res.body.encodedPayload === '');
+  });
+
+  // if (
```
enddynayn, line 110: left over code?

apps/account-worker/src/transaction_publisher/publisher.service.ts
```diff
@@ -196,6 +204,34 @@ export class TransactionPublisherService extends BaseConsumer implements OnAppli
...
+    try {
+      const { nonce } = await this.blockchainService.api.query.system.account(accountId);
+
+      this.logger.debug(signer, 'signer****');
+
+      const submittableExtrinsic = await ext.signAsync(accountId, { nonce, signer });
+      this.logger.debug('MADE IT');
```
enddynayn, line 219: left over stuff?

enddynayn:
It’s coming together! The tough part is done, and we’re almost there. Just a little more to go —great work so far! 👍

apps/account-api/src/controllers/v1/accounts-v1.controller.ts
```diff
@@ -101,4 +104,45 @@ export class AccountsControllerV1 {
...
+    } catch (error) {
+      this.logger.error(error);
+      throw new HttpException('Failed to create a retireMsa unsigned, encoded extrinsic.', HttpStatus.BAD_REQUEST);
+    }
+  }
+
+  @Post('retireMsa')
+  @HttpCode(HttpStatus.CREATED)
+  @ApiOperation({ summary: 'Request to retire an Msa Id.' })
+  @ApiCreatedResponse({ description: 'Signed in successfully', type: TransactionResponse })
```
mattheworris, line 132:
```suggestion
  @ApiCreatedResponse({ description: 'Created and queued request to retire an Msa Id', type: TransactionResponse })
```
```suggestion
  @ApiCreatedResponse({ description: 'Signed in successfully', type: TransactionResponse })
```

apps/account-worker/src/transaction_publisher/publisher.service.ts
```diff
@@ -196,6 +203,26 @@ export class TransactionPublisherService extends BaseConsumer implements OnAppli
     }
   }
 
+  async processProxyTxn(
+    ext: SubmittableExtrinsic<'promise', ISubmittableResult>,
+    accountId: string,
+    signer: Signer,
+  ): Promise<[SubmittableExtrinsic<'promise'>, HexString]> {
+    try {
+      const { nonce } = await this.blockchainService.api.query.system.account(accountId);
```
aramikm, line 212: Should this use `nonceService`? We might need to add a new method in there that gets accountId as a param

apps/account-api/src/controllers/v1/accounts-v1.controller.ts
```diff
@@ -101,4 +104,47 @@ export class AccountsControllerV1 {
...
+  /**
+   * Posts the signer and payload, signing the retire msa payload and executing that extrinsic.
+   * @returns Returns a TransactionResponse hex string when the extrinsic is added to the queue.
+   * @throws An error if the signed extrinsic fails to be created.
+   */
+  async postRetireMsa(@Body() retireMsaRequest: RetireMsaRequestDto): Promise<TransactionResponse> {
+    try {
+      return this.accountsService.retireMsa(retireMsaRequest);
+    } catch (error) {
+      const errorMessage = 'Failed to Sign In With Frequency';
```
mattheworris, line 145:
```suggestion
      const errorMessage = 'Failed to create request to retire an Msa Id';
```

apps/account-api/test/accounts.controller.e2e-spec.ts
claireolmstead: resolved

apps/account-worker/src/transaction_publisher/publisher.service.ts
claireolmstead: Decided that this was out of scope.

apps/account-api/src/controllers/v1/accounts-v1.controller.ts
```diff
@@ -119,7 +119,9 @@ export class AccountsControllerV1 {
     @Param('accountId') accountId: string,
   ): Promise<{ signerPayload: SignerPayloadRaw; encodedPayload: string }> {
```
JoeCap08055, line 120: This should be a defined type somewhere; I see it's used in other places, shouldn't have to write it out everywhere we use it. Also a common type definition gives us compile-time type-safety checks.

apps/account-api/test/accounts.controller.e2e-spec.ts
```diff
@@ -100,35 +102,53 @@ describe('Account Controller', () => {
...
-    console.log('getRetireMsaResponse:', getRetireMsaResponse.body.data);
-    const responseData = getRetireMsaResponse.body.data;
-    // To be removed
-    await cryptoWaitReady();
-
-    const uint8Signature = keypair.sign(JSON.stringify(responseData.unsignedPayload), { withType: false });
-    const signature = u8aToHex(uint8Signature);
+    const getPath: string = `/v1/accounts/retireMsa/${accountId}`;
+    const getRetireMsaResponse = await request(app.getHttpServer()).get(getPath);
+    console.log('getRetireMsaResponse:', getRetireMsaResponse.body);
```
JoeCap08055, line 133:
```suggestion
```

apps/account-api/test/accounts.controller.e2e-spec.ts
```diff
@@ -100,35 +102,53 @@ describe('Account Controller', () => {
...
+    const responseData: { signerPayload: SignerPayloadRaw; encodedPayload: string } = getRetireMsaResponse.body;
```
JoeCap08055, line 134: Use a common type def here

apps/account-api/src/controllers/v1/accounts-v1.controller.ts
claireolmstead: Meant to do that and forgot in everything else. thanks for the catch.

apps/account-api/test/accounts.controller.e2e-spec.ts
```diff
@@ -100,35 +102,53 @@ describe('Account Controller', () => {
...
 
-    const retireMsaRequest = {
-      unsignedPayload: responseData.unsignedPayload,
+    await cryptoWaitReady();
+    const tx = ExtrinsicHelper.apiPromise.tx.msa.retireMsa();
+    const signerPayload: SignerPayloadRaw = await getRawPayloadForSigning(tx, accountId);
+    const { data } = signerPayload;
+    const signature: Uint8Array = keypair.sign(data, { withType: true });
+    const prefixedSignature: SignerResult = { id: 1, signature: u8aToHex(signature) };
+    const signer: Signer = getSignerForRawSignature(prefixedSignature);
```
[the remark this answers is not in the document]
claireolmstead: In the test, we need to generate the signer that is used to sign the payload. Per my understanding, in a real world scenario, that code would be executed in a separate UI and returned.

libs/account-lib/src/blockchain/blockchain.service.ts
```diff
@@ -511,4 +511,52 @@ export class BlockchainService implements OnApplicationBootstrap, OnApplicationS
...
+    tx: SubmittableExtrinsic<'promise', ISubmittableResult>,
+    signerAddress: string,
+    // eslint-disable-next-line consistent-return
+  ): Promise<SignerPayloadRaw> {
+    let signRaw;
+    try {
+      await tx.signAsync(signerAddress, {
+        signer: {
+          signRaw: (raw) => {
+            console.log('signRaw called with [raw]:', raw);
```
mattheworris, line 526:
```suggestion
            this.logger.verbose('signRaw called with [raw]:', raw);
```

JoeCap08055 (approved):

apps/account-api/test/accounts.controller.e2e-spec.ts
```diff
@@ -161,4 +165,58 @@ describe('Account Controller', () => {
...
+      };
+      const expectedEncodedDataResult: string =
+        '0x3c0ac40000005e0000000100000085f854538489ccf7ebbc59e571bf44a207fd007505db84234a51120688e44f0bb0ffe32b3a6e80195781b56bb94ae41c9c7fadc4ecf27bc488c32dfac3c3b31c';
+
+      await request(app.getHttpServer())
+        .get(path)
+        .expect(200)
+        .expect((res) => res.body.signerPayload === expectedSignerPayloadResult)
+        .expect((res) => res.body.encodedPayload === expectedEncodedDataResult);
+    });
```
mattheworris, line 187: I think this needs to be updated for the new dto structure?

apps/account-api/test/e2e-setup.mock.spec.ts
```diff
@@ -85,3 +87,39 @@ export async function generateAddPublicKeyExtrinsic(
...
+    });
+  } catch (e: any) {
+    // If we encountered an actual error, re-throw it; otherwise
+    // ignore the fake error we threw above
+    if (e?.message !== fakeError) {
+      throw e;
+    }
+  }
+  return signRaw;
+};
```
mattheworris, line 125: This should not be needed, if the test above is refactored.
```suggestion
```

mattheworris (approved):
- Read through the changes.
- I think the e2e tests are not up to date, but otherwise looks good.
🚢 it!

PR #488, opened 2024-09-06
account-service: added graph key endpoints

### LINKED ISSUE #414: `applyItemActions` (for graph encryption key management)
# Description
As a Provider/consumer of the Gateway API, I want to be able to manage encryption keys for a user (for Private Graph, or any other purpose). This particular endpoint should be specific to the dsnp private graph encryption key

See @claireolmstead's prs for a pattern

# Discussion
- Which version(s) of this extrinsic to support--signed, delegated, or both?
    - Signed only
- Does this belong in `account` or `graph` service?
    - `account` service

aramikm:
# Problem
To be able to easily add graph keys we need endpoints which facilitates this operation.

# Solution
implemented two new endpoints which
- Creates an encoded payload which is ready for signing
- Publishes a new graph key on chain
closes #414

## Screenshots (optional):

- new endpoints
Screenshot 2024-09-09 at 2 06 05 PM

- endpoint that prepared the payload to add a new graph key
Screenshot 2024-09-09 at 2 07 00 PM

- endpoint that submits the graph key creation

Screenshot 2024-09-09 at 2 07 16 PM

- worker logs
Screenshot 2024-09-06 at 2 12 01 PM

- blockchain submission
Screenshot 2024-09-06 at 2 12 27 PM

Screenshot 2024-09-06 at 2 13 47 PM


apps/account-worker/src/transaction_notifier/notifier.service.ts
```diff
@@ -187,6 +187,19 @@ export class TxnNotifierService
                 );
               }
               break;
+            case TransactionType.ADD_GRAPH_KEY:
```
aramikm, line 190: What if we don't want to notify via web-hooks? Doesn't seems like there are any options to disable it

mattheworris (approved):
- Read through changes, looks good!
- Do we want to add tests for these endpoints?

🚢 it!

apps/account-api/src/controllers/v1/graphs-v1.controller.ts
JoeCap08055, line 1: I'm not sure this needs to be its own controller, for 2 reasons:
1. It's really key-management, not graph-management
2. Currently, we only support 2 kinds of keys: account control keys, and graph encryption keys. But in future there may be other types of keys to support in additional `Itemized` schemas, using the same pattern as graph keys, simply with a different target schema and (potentially) a different key type. It would probably make sense to group this all under the same controller. Perhaps the controller endpoints could be grouped under the `/keys` controller like so:
- `/keys/<version>/<existing keys endpoints>`
- `/keys/<version>/graphKeys/<graph key endpoints>`

or, if we think we might want to version the control key endpoints separately from the `Itemized` storage key endpoints, perhaps rename the existing `keys` controller to `/controlKeys`, and rename the new controller  to something (not using 'graph', perhaps `/publicKeyAgreements` or something?)

apps/account-api/src/services/graphs.service.ts
JoeCap08055, line 1: Perhaps `GraphKeyService` would be a better name?

This also makes me think that perhaps the key-management piece of `graph-sdk` should be split out as its own library... future discussion

apps/account-api/src/services/graphs.service.ts
```diff
@@ -0,0 +1,78 @@
...
+    this.logger = new Logger(this.constructor.name);
+    const environmentType = this.configService.graphEnvironmentType;
+    const environment: EnvironmentInterface = { environmentType: EnvironmentType[environmentType] };
+    const graphState = new Graph(environment);
+    // there might be a better way to get this schema id but for now we are stuck to get it from graph-sdk
+    this.graphKeySchemaId = graphState.getGraphConfig(environment).graphPublicKeySchemaId;
+    graphState.freeGraphState();
+  }
+
+  async getAddingNewKeyPayload(msaId: string, newKey: HexString): Promise<AddNewGraphKeyPayloadRequest> {
```
JoeCap08055, line 33:
```suggestion
  async getAddNewKeyPayload(msaId: string, newKey: HexString): Promise<AddNewGraphKeyPayloadRequest> {
```

apps/account-api/src/services/graphs.service.ts
```diff
@@ -0,0 +1,78 @@
...
+    try {
+      const expiration = await this.getExpiration();
+      const schemaId = this.graphKeySchemaId;
+      const itemizedStorage = await this.blockchainService.getItemizedStorage(msaId, schemaId);
+      if (
+        itemizedStorage.items
+          .toArray()
+          .find((i) => i.payload.length > 0 && u8aToHex(i.payload).toLowerCase() === newKey.toLowerCase())
+      ) {
+        throw new BadRequestException('Requested key already exists!');
```
JoeCap08055, line 43: Seems like it's recommended to return `HTTP 409 CONFLICT` when requested to create duplicate content

apps/account-worker/src/transaction_notifier/notifier.service.ts
JoeCap08055: I would ultimately like to standardize across all Gateway services the following behavior for webhooks:
* clients of certain services (where appropriate) may register a global webhook to be called for all responses requiring a notification
* all requests initiating a delayed processing response may include an optional URL field for a request-specific webhook to be propagated with the request job
* completed requests will notify both the global & request-specific webhooks, if specified. If neither registered, no webhook notification will be attempted & it's not an error.
* (optional; to discuss) all services should provide a request polling endpoint for clients that prefer to poll (this, however, requires a bit more state caching in the service than is currently implemented)

apps/account-worker/src/transaction_publisher/publisher.service.ts
```diff
@@ -114,8 +124,10 @@ export class TransactionPublisherService extends BaseConsumer implements OnAppli
         providerId: job.data.providerId,
         txHash,
         successEvent: targetEvent,
-        birth: tx.era.asMortalEra.birth(lastFinalizedBlockNumber),
-        death: tx.era.asMortalEra.death(lastFinalizedBlockNumber),
+        birth: tx.era.isMortalEra ? tx.era.asMortalEra.birth(lastFinalizedBlockNumber) : lastFinalizedBlockNumber,
+        death: tx.era.isMortalEra
+          ? tx.era.asMortalEra.death(lastFinalizedBlockNumber)
+          : lastFinalizedBlockNumber + IMMORTAL_BLOCK_PERIOD,
```
JoeCap08055, line 130: No transactions should be immortal, and I believe I've found the issue. You're trying to get the mortality era of the wrapped transaction, when you really need the era of the `frequencyTxnPayment.payWithCapacity` transaction. `processSingleTransaction` needs to be modified to return the same as `processBatchTransaction`; that is, instead of returning just the hash, it needs to return `Promise<[SubmittableExtrinsic<'promise'>, HexString]>`. See the example of `SIWF_SIGNUP` above for how this works.

jest.config.json
JoeCap08055, line 1: Should we add this to `.prettierignore`?

openapi-specs/account.openapi.json
JoeCap08055, line 1: I think this also should be in `.pretttierignore`

openapi-specs/content-watcher.openapi.json
JoeCap08055, line 1: .prettierignore

openapi-specs/graph.openapi.json
JoeCap08055, line 1: .prettierignore

JoeCap08055:
Overall, looks pretty good. Left some comments inline. My biggest comment is for naming REST API endpoints. I've been trying to follow [this approach](https://restfulapi.net/resource-naming/). Going by that, perhaps the following endpoint names would be better, or something similar:

```
GET <path>/graphKeys/addKeyPayload/{key} // see my inline comment about this one; it may need to be a POST
// or
POST <path>/graphKeys/getAddKeyPayload

POST <path>/graphKeys // no need for any other specifier; the HTTP method & resource name tell us what the endpoint does

GET <path>/graphKeys/{msaId} // return list of all graph keys for MSA
GET <path>/graphKeys/{msaId}/{index} // get specific graph key from MSA's list of graph keys
```

apps/account-api/src/metadata.ts
```diff
@@ -1,14 +1,16 @@
 /* eslint-disable */
 export default async () => {
     const t = {
```
claireolmstead, line 3: make sure to run format. this looks unformatted.

libs/account-lib/src/blockchain/blockchain.service.ts
```diff
@@ -421,6 +479,26 @@ export class BlockchainService implements OnApplicationBootstrap, OnApplicationS
...
+    if (event && this.api.events.statefulStorage.ItemizedPageUpdated.is(event)) {
+      graphKeyValues.msaId = event.data.msaId.toString();
+      graphKeyValues.schemaId = event.data.schemaId.toString();
+      graphKeyValues.prevContentHash = event.data.prevContentHash.toString();
+      graphKeyValues.currContentHash = event.data.currContentHash.toString();
+      graphKeyValues.debugMsg = `New Graph Public Key Added for msaId: ${graphKeyValues.msaId} and schemaId: ${graphKeyValues.schemaId}`;
+    }
+
+    return graphKeyValues as ItemizedPageUpdated;
+  }
```
claireolmstead, line 500:
```suggestion
  public handlePublishGraphKeyTxResult(event: Event): ItemizedPageUpdated {
    // Grab the event data
    if (event && this.api.events.statefulStorage.ItemizedPageUpdated.is(event)) {
      return {
        msaId: event.data.msaId.toString(),
        schemaId: event.data.schemaId.toString(),
        prevContentHash: event.data.prevContentHash.toString(),
        currContentHash: event.data.currContentHash.toString(),
        debugMsg: `New Graph Public Key Added for msaId: ${event.data.msaId} and schemaId: ${event.data.schemaId}`,
      };
    }
    throw new Error('Graph Key Tx Result Event invalid.');
  }
```

apps/account-worker/src/transaction_publisher/publisher.service.ts
aramikm: Good find. Applied the changes and that solved the issue. Thanks

apps/account-api/src/services/graphs.service.ts
aramikm: moved into key-service.

apps/account-api/src/controllers/v1/keys-v1.controller.ts
```diff
@@ -76,4 +85,56 @@ export class KeysControllerV1 {
...
   }
+
+  @Get('publicKeyAgreements/getAddKeyPayload')
+  @HttpCode(HttpStatus.OK)
+  @ApiQuery({
+    name: 'newKey',
+    description: 'New public key to be added in hex format',
+    type: 'string',
+    required: true,
+  })
```
JoeCap08055, line 96: Don't need this decorator if `@Query()` parameter is a properly-decorated DTO class (see other comment)

apps/account-api/src/controllers/v1/keys-v1.controller.ts
```diff
@@ -76,4 +85,56 @@ export class KeysControllerV1 {
...
+   * @throws An error if the key already exists or the payload creation fails.
+   */
+  async getPublicKeyAgreementsKeyPayload(
+    @Query('msaId') msaId: string,
+    @Query('newKey') newKey: HexString,
+  ): Promise<AddNewPublicKeyAgreementPayloadRequest> {
+    // this is temporary until we find a better way to enforce data validation. the validation decorator didn't work
+    if (!isHexString(newKey)) {
+      throw new BadRequestException('Not a valid Hex value!');
+    }
```
JoeCap08055, line 112: Validation decorators can work on query parameters if you create a DTO for the `@Query()` parameter and decorate it appropriately. Something like:

```
class AddNewKeyPayloadRequestDto {
  @IsNumberString()
  @ApiProperty({ type: String, description: 'MSA ID of account to generate a payload for' })
  msaId: string;

  @IsHexadecimal()
  @ApiProperty({ type: String, description: 'New public key to be added in hex format' })
  newKey: HexString;
}
```

Then in your controller method signature, instead of individual query parameters, you have a single parameter like so:
```
  async getPublicKeyAgreementsKeyPayload(
    @Query() { msaId, newKey }: AddNewKeyPayloadRequestDto,
  ): Promise<AddNewPublicKeyAgreementPayloadRequest> {

```

libs/account-lib/src/utils/custom.decorator.ts
```diff
@@ -0,0 +1,25 @@
+import { registerDecorator, ValidationArguments, ValidationOptions } from 'class-validator';
+
+export function IsHexPublicKey(validationOptions?: ValidationOptions) {
```
JoeCap08055, line 3: might be nice to add some `validationOptions` in here, such as:

* `prefixRequired: boolean`
* `minByteLen: number`
* `maxByteLen: number`

Would make it easy to support different key algorithms in the future.

aramikm: I'll add this in the validation specific ticket.

As an unrelated note I think going forward we have to be more careful about adding new features and functionality in general. We have a lot of existing features that needs to be hardened and adding more would only make our life harder.

One question I usually like to think about is that how hard would it be for a future developer to change this to support that extra feature or data. If it is easy enough and the probability of this being changed in the near future is low enough. Usually keeping it simple and dumb might be preferable.

PR #502, opened 2024-09-10
chore: various fixes to get e2e tests working for account service

mattheworris:
# Problem

After changing to the mono-repo, e2e tests do not work because the scripts that execute them need updates, due to file locations changing.

# Solution

Update the e2e test workflow so that the tests can execute locally.

## Steps to Verify:

1. Follow the steps in `./developer-docs]account/README.md` and verify that the tests execute correctly.


Makefile
```diff
@@ -54,8 +54,8 @@ $(SWAGGER_TARGETS): swagger-%: openapi-%
 $(TEST_TARGETS):
 	@npm run test:$(@:test-%=%)
 
-# $(E2E_TARGETS):
-# 	@( cd apps/$(@:test-e2e-%=%) ; npm run test:e2e )
+$(E2E_TARGETS):
+	@npm run test:e2e:$(@:test-e2e-%=%)
```
mattheworris, line 58: Allows `make test-e2e-account` to work correctly.

apps/account-api/test/accounts.controller.e2e-spec.ts
```diff
@@ -82,8 +82,8 @@ describe('Account Controller', () => {
     const invalidMsaId = BigInt(maxMsaId) + 1000n;
     await request(app.getHttpServer())
       .get(`/v1/accounts/${invalidMsaId.toString()}`)
-      .expect(400)
-      .expect({ statusCode: 400, message: 'Failed to find the account' });
+      .expect(404)
```
mattheworris, line 85: Updating for changes to statusCode.

developer-docs/account/README.md
```diff
@@ -206,23 +206,23 @@ Note: using [docker compose file](../../docker-compose.yaml) to start the servic
 The following command will start all of the necessary containers for the account service to run in development mode.
 
 ```bash
-make restart-chain-docker
+./scripts/account/restart-chain-docker
```
mattheworris, line 209: Opted to specify the scripts that are specific to account service, rather than propagate a make command, which would need to be duplicated for each service app.

developer-docs/account/README.md
```diff
@@ -254,11 +254,13 @@ docker compose up -d account-service-api account-service-worker
 
 ### Run the tests
 
+Make sure to follow the steps above and use `restart-chain-docker.sh` to start the services. This will ensure that the local node is running and is properly configured. Then run the following command to run the tests. **NOTE:** The tests are not idempotent. When re-running the e2e tests, use the `restart-chain-docker.sh` script to reset the chain.
+
 ```bash
-make test-e2e
+make test-e2e-account
```
mattheworris, line 260: This pattern already existed and works.

scripts/account/docker-clean-up.sh
```diff
@@ -8,8 +8,11 @@ docker compose down frequency redis account-service-api account-service-worker
 echo "Removing specified volumes..."
 docker volume rm gateway_redis_data
 docker volume rm gateway_chainstorage
-docker volume rm gateway_account_node_cache
+docker volume rm gateway_account_api_node_cache
+docker volume rm gateway_account_worker_node_cache
+docker volume prune -f
```
JoeCap08055, line 13: nit: let's use `-f` flag here so the script doesn't complain if any of the volumes don't exist. (I checked and that's all the flag does)

scripts/account/restart-chain-docker.sh
JoeCap08055, line 1: Can we also put the `mock-webhook-server` in a container & add to docker-compose? Will be easier when we want to run this all in the pipeline.

JoeCap08055 (commented):
WIth some pretty minor changes, copying & modifying the `load-tests.yml` CI workflow should get this running in CI pretty easily, especially if you containerize the webhook server.

aramikm:
I'm seeing the following error when I run it using  `make test-e2e-account` command

Unable to transfer initial token amount 1000000:
                RpcError: 1014: Priority is too low: (1032 vs 1032): The transaction has too low priority to replace another transaction already in the pool.

JoeCap08055 (approved):

apps/account-api/test/handles.controller.e2e-spec.ts
```diff
@@ -61,23 +64,37 @@ describe('Handles Controller', () => {
     }).compile();
 
     app = module.createNestApplication();
+    app.enableShutdownHooks();
+    app.useGlobalPipes();
```
mattheworris, line 68: Are these redundant with lines 74-75?
```suggestion
```

.github/workflows/load-tests.yml
```diff
@@ -78,25 +78,12 @@ jobs:
...
         run: npm ci
 
-      - name: Set up Docker Buildx
-        if: ${{ steps.should.outputs.RUN == 'true' }}
-        uses: docker/setup-buildx-action@v3
-        # Use GitHub Container Registry instead due to rate limits
-        with:
-          buildkitd-config-inline: |
-            [registry."ghcr.io"]
-      - name: Login to DockerHub
```
wilwade, line 88: We still need to setup to use ghcr don't we? Otherwise we'll hit the rate limits of docker hub again right?

mattheworris (commented):
🚢 it!

.github/workflows/load-tests.yml
JoeCap08055: Ah, okay. But do we need to login? We're not pushing anything.

wilwade: No. As long as we are pulling from ghcr, we don't need to.

PR #514, opened 2024-09-12
Implement revokeDelegation(ByDelegator|ByProvider)

### LINKED ISSUE #411: `revokeDelegation(ByDelegator|ByProvider)`
# Description
As a consumer of the Gateway API, I want to be able to revoke a previously granted delegation

Should implement:
`GET /account/accounts/revokeDelegation/{providerId}` (returns a `revokeDelegationByDelegator` encoded extrinsic that can be signed)
`POST /account/accounts/revokeDelegation` submits a signed extrinsic for execution

mattheworris:
# Description
As a consumer of the Gateway API, I want to be able to revoke a previously granted delegation

Should implement:

- [x] `GET /account/accounts/revokeDelegation/{accountId}/{providerId}` 
  - [x] (returns a `revokeDelegationByDelegator` encoded extrinsic that can be signed)
- [x] `POST /account/accounts/revokeDelegation` submits a signed extrinsic for execution

Closes #411 

# Solution

New endpoints were added to implement the desired functionality.

## Steps to Verify:

1. Execute `./scripts/account/restart-chain-docker.sh` and make sure `account-service` is running. You will need to additionally run the following command to start the account service, or start it locally, as the restart-chain script no longer brings up account.
  - `docker compose -f docker-compose.yaml -f docker-compose-e2e.account.yaml --profile e2e --profile account up -d`
3. Execute `cd ./apps/account-api/test/setup/ && npm run test-revoke-e2e`
4. Verify that `msa.DelegationRevoked` event appears on the local node.
Note: This will only work once and then the chain state needs to be restarted.


.markdownlint.json
JoeCap08055, line 1: Should this be here?

apps/account-api/src/controllers/v1/delegation-v1.controller.ts
```diff
@@ -28,4 +50,51 @@ export class DelegationControllerV1 {
       throw new HttpException('Failed to find the delegation', HttpStatus.BAD_REQUEST);
     }
   }
+
+  @Get('revokeDelegation/:accountId/:providerId')
+  @HttpCode(HttpStatus.OK)
+  @ApiOperation({ summary: 'Get a properly encoded RevokeDelegationPayload that can be signed' })
+  @ApiOkResponse({ description: 'Returned an encoded RevokeDelegationPayload for signing' })
```
JoeCap08055, line 57: add `type: RevokeDelegationPayloadResponseDto` to `ApiOkResponse()` so that the response payload is included in the generated OpenAPI spec.

.markdownlint.json
mattheworris: I would like it to be, but it is probably specific to my vscode extension for markdown, so will take it out.

apps/account-api/src/controllers/v1/delegation-v1.controller.ts
```diff
@@ -28,4 +50,51 @@ export class DelegationControllerV1 {
...
+      this.logger.error(error);
+      throw new HttpException('Failed to generate the RevokeDelegationPayload', HttpStatus.BAD_REQUEST);
+    }
+  }
+
+  // Pass through the revoke delegation request to the blockchain
+  @Post('revokeDelegation')
+  @HttpCode(HttpStatus.CREATED)
+  @ApiOperation({ summary: 'Request to revoke a delegation' })
+  @ApiCreatedResponse({ description: 'Created and queued request to revoke a delegation' })
```
JoeCap08055, line 86: add `{ type: TransactionResponse }` for OpenAPI response payload

apps/account-api/src/services/delegation.service.ts
```diff
@@ -36,4 +47,23 @@ export class DelegationService {
     }
     throw new Error('Invalid msaId.');
   }
+
+  async getRevokeDelegationPayload(accountId: string, providerId: string): Promise<RevokeDelegationPayloadResponseDto> {
+    // TODO: Validate input: make sure accountId and providerId are valid and the delegation exists
```
JoeCap08055, line 52: Finish TODO

JoeCap08055: I recommend the following responses:

* either `accountId` or `providerId` does not exist as an MSA => `HttpStatus.NOT_FOUND`
* `providerId` is a valid MSA but not a registered provider => `HttpStatus.BAD_REQUEST`
* no delegation found for account/provider pair => `HttpStatus.NOT_FOUND`

apps/account-api/test/delegations.controller.e2e-spec.ts
```diff
@@ -101,4 +103,29 @@ describe('Delegation Controller', () => {
         revokedAt: '0x00000000',
       });
   });
+
+  it('(POST) /v1/delegation/revokeDelegation/:accountId/:providerId', async () => {
+    const providerId = provider.msaId?.toString();
+    const { keypair } = users[1];
+    const accountId = keypair.address;
+    const getPath: string = `/v1/delegation/revokeDelegation/${accountId}/${providerId}`;
+    const response = await request(app.getHttpServer()).get(getPath).expect(200);
```
JoeCap08055, line 112: Could either:
* add a separate test for `(GET) /v1/delegation/revokeDelegation/:accountId/:providerId` specifically, -OR-
* add some `expect` statements here to validate the response payload, and continue on to POST

apps/account-api/test/setup/package.json
```diff
@@ -5,6 +5,8 @@
   "main": "index.ts",
   "scripts": {
     "main": "tsx index.ts",
+    "test-revoke": "tsx test-revoke.ts",
+    "test-revoke-e2e": "tsx test-revoke-e2e.ts",
```
JoeCap08055, line 9: Do we need these?
Maybe move `test-revoke.ts` to the Lore repo, although the one for `retireMsa` is already there; same flow, different extrinsic.

apps/account-api/test/setup/test-revoke-e2e.ts
JoeCap08055, line 1: This seems redundant to the E2E test

apps/account-api/test/setup/test-revoke.ts
JoeCap08055, line 1: Remove?

docker-compose-e2e.account.yaml
```diff
@@ -32,13 +32,16 @@ services:
...
 
   frequency:
+    image: dsnp/instant-seal-node-with-deployed-schemas:latest
+    # We need to specify the platform because it's the only image
+    # built by Frequency at the moment, and auto-pull won't work otherwise
+    platform: linux/amd64
+    # SEALING_MODE set to insstant for e2e testing
+    environment:
+      - SEALING_MODE=interval
+      - SEALING_INTERVAL=1
```
JoeCap08055, line 42: Shouldn't need to add this here; don't we always layer this file in *after* the main `docker-compose.yaml`, which already specifies this?

docker-compose-e2e.account.yaml
```diff
@@ -32,13 +32,16 @@ services:
...
     profiles:
       - e2e
-      - local-node
-
-  mock-webhook-server:
-    profiles:
-      - skip
```
JoeCap08055, line 41: Why remove?

libs/account-lib/src/types/dtos/revokeDelegation.request.dto.ts
```diff
@@ -0,0 +1,40 @@
...
+  providerId: string;
+}
+
+export class RevokeDelegationPayloadResponseDto {
+  @ApiProperty()
+  @IsNotEmpty()
+  accountId: string;
+
+  @ApiProperty()
+  @IsNotEmpty()
```
JoeCap08055, line 20: Consider `@IsNumberString()`

JoeCap08055: (normally don't need validators on a response type, but since the request extends this class, it's good to have)

libs/account-lib/src/types/dtos/revokeDelegation.request.dto.ts
```diff
@@ -0,0 +1,40 @@
...
+  providerId: string;
+
+  @ApiProperty()
+  @IsNotEmpty()
+  encodedExtrinsic: HexString;
+
+  @ApiProperty()
+  @IsNotEmpty()
```
JoeCap08055, line 28: Consider `@IsHexadecimal()`

openapi-specs/account.openapi.json
```diff
@@ -139,6 +139,63 @@
...
+            "in": "path",
+            "schema": {
+              "type": "string"
+            }
+          }
+        ],
+        "responses": {
+          "200": {
+            "description": "Returned an encoded RevokeDelegationPayload for signing"
+          }
```
JoeCap08055, line 167: note this response type is empty; see previous comments about adding `type` annotations in the decorators.

apps/account-worker/src/transaction_publisher/publisher.service.ts
```diff
@@ -196,6 +204,28 @@ export class TransactionPublisherService extends BaseConsumer implements OnAppli
     }
   }
 
+  async processProxyTxn(
+    ext: SubmittableExtrinsic<'promise', ISubmittableResult>,
+    accountId: string,
+    signature: HexString,
+  ): Promise<[SubmittableExtrinsic<'promise'>, HexString]> {
+    try {
+      const prefixedSignature: SignerResult = { id: 1, signature };
```
aramikm, line 213: is this the same as `SpRuntimeMultiSignature` ?

If that is the case then one other way we created these were like the following example

```rust
const proof = { Sr25519: u8aToHex(delegatorKeys.sign(u8aWrapBytes(payload.toU8a()))) };
```

libs/account-lib/src/types/dtos/revokeDelegation.request.dto.ts
aramikm: I'll fix all these, I would suggest not spending too much on these.

apps/account-api/test/setup/test-revoke-e2e.ts
mattheworris: This runs against the full service and confirms that the transaction exectues, which doesn't happen for the jest based e2e tests.

apps/account-api/test/setup/package.json
JoeCap08055: But isn't that e2e logic already in `delegations.controller.e2e-spec.ts`?

apps/account-api/test/setup/test-revoke-e2e.ts
JoeCap08055: I don't see where it confirms that the transaction executes; it just calls the POST endpoint and returns at the end... I agree that's what our E2E tests _should_ do eventually, but I think we're going to take another issue for that overhaul.

mattheworris: Clarification: the transaction *can be* verified on the chain. No, the code doesn't check for it.

libs/account-lib/src/types/dtos/accounts.request.dto.ts
```diff
@@ -1,7 +1,7 @@
 // eslint-disable-next-line max-classes-per-file
 import { ApiProperty } from '@nestjs/swagger';
 import { IsHexadecimal, IsNotEmpty } from 'class-validator';
-import { TransactionType } from '#account-lib';
+import { TransactionType } from '../enums';
```
mattheworris, line 4: This caused a Dependency cycle. I think this is related to the indexing problems we saw previously when importing and the services were being included when not needed.

libs/account-lib/src/types/dtos/accounts.request.dto.ts
```diff
@@ -1,7 +1,7 @@
 // eslint-disable-next-line max-classes-per-file
 import { ApiProperty } from '@nestjs/swagger';
 import { IsHexadecimal, IsNotEmpty } from 'class-validator';
-import { TransactionType } from '#account-lib';
+import { TransactionType } from '#account-lib/types/enums';
```
mattheworris, line 4: This caused a Dependency cycle. I think this is related to the indexing problems we saw previously when importing and the services were being included when not needed.

aramikm (approved):
Looks good to me!

PR #515, opened 2024-09-13
graph-api: added more validation

### LINKED ISSUE #496: graph-api endpoints data syntax verification
Checking type, length, format, ...

aramikm:
# Problem

Adds extra validation on graph-api dtos and request/path params

Closes #496 

# Discussion
- Since we are going to overhaul our way of e2e testing it didn't make sense to add tests which would be replaced in the future. 

## Steps to Verify:

1. Run the api
2. Go to swagger
3. Try some invalid values in requests


tsconfig.json
```diff
@@ -38,7 +38,9 @@
       "#graph-lib": ["libs/graph-lib/src"],
       "#graph-lib/*": ["libs/graph-lib/src/*"],
       "#types": ["libs/types/src"],
-      "#types/*": ["libs/types/src/*"]
+      "#types/*": ["libs/types/src/*"],
+      "#utils": ["libs/utils/src"],
```
aramikm, line 42: This is a new directory to store the common utils liek custom guards or decorators. please let me know if you have a better idea how to separate and keep these.

apps/graph-api/src/guards/read-only-deployment-guard.service.ts
```diff
@@ -2,7 +2,7 @@ import { ConfigService } from '#graph-lib/config';
 import { Injectable, CanActivate, ExecutionContext } from '@nestjs/common';
 
 @Injectable()
-export class ReadOnlyGuard implements CanActivate {
+export class ReadOnlyDeploymentGuard implements CanActivate {
```
aramikm, line 5: We have this Guard duplicated in other applications but I didn't want to increase the size of this PR but those will be also replaced and refactored in the future.

libs/types/src/dtos/graph/watch-graphs.dto.ts
```diff
@@ -15,7 +16,7 @@ export class WatchGraphsDto {
 
   @IsNotEmpty()
   @IsString()
-  @IsUrl({ require_tld: false })
+  @IsUrl({ require_tld: false, require_protocol: true, require_valid_protocol: true })
```
saraswatpuneet, line 19: These are good to know 💡

saraswatpuneet (approved):
👍🏽

apps/graph-api/src/guards/read-only-deployment-guard.service.ts
saraswatpuneet: Maybe we can add it to some existing story as a comment

apps/graph-api/src/controllers/v1/webhooks-v1.controller.ts
```diff
@@ -51,8 +52,8 @@ export class WebhooksControllerV1 {
     required: false,
     description: "Boolean whether to include webhooks registered for 'all' MSA Ids (default: true)",
   })
-  async getWebhooksForMsa(@Param('msaId') msaId: string, @Query('includeAll') includeAll = 'true'): Promise<string[]> {
-    return this.apiService.getWebhooksForMsa(msaId, JSON.parse(includeAll ?? 'true'));
+  async getWebhooksForMsa(@Param() msaId: MsaIdParam, @Query('includeAll') includeAll = 'true'): Promise<string[]> {
+    return this.apiService.getWebhooksForMsa(msaId.msaId, JSON.parse(includeAll ?? 'true'));
```
JoeCap08055, line 56:
```suggestion
  async getWebhooksForMsa(@Param() { msaId }: MsaIdParam, @Query('includeAll') includeAll = 'true'): Promise<string[]> {
    return this.apiService.getWebhooksForMsa(msaId, JSON.parse(includeAll ?? 'true'));
```

apps/graph-api/src/controllers/v1/webhooks-v1.controller.ts
```diff
@@ -51,8 +52,8 @@ export class WebhooksControllerV1 {
     required: false,
     description: "Boolean whether to include webhooks registered for 'all' MSA Ids (default: true)",
   })
-  async getWebhooksForMsa(@Param('msaId') msaId: string, @Query('includeAll') includeAll = 'true'): Promise<string[]> {
-    return this.apiService.getWebhooksForMsa(msaId, JSON.parse(includeAll ?? 'true'));
+  async getWebhooksForMsa(@Param() msaId: MsaIdParam, @Query('includeAll') includeAll = 'true'): Promise<string[]> {
```
JoeCap08055, line 55: I forget how we standardized parameter naming, but I _think_ we landed on suffixing REST input parameter object types with `Dto` or `DTO`.

Also, I would like to try and standardize on documenting the DTO classes with `@ApiProperty()` decorators, which eliminates the need to document each endpoint with `@ApiParameter()` or `@ApiQuery()`

apps/graph-api/src/controllers/v1/webhooks-v1.controller.ts
```diff
@@ -65,8 +66,8 @@ export class WebhooksControllerV1 {
     example: 'http://localhost/webhook',
     description: 'URL for which to fetch all watched MSAs',
   })
-  async getWebhooksForUrl(@Query('url') url: string): Promise<string[]> {
-    return this.apiService.getWatchedGraphsForUrl(url);
+  async getWebhooksForUrl(@Query() url: UrlParam): Promise<string[]> {
+    return this.apiService.getWatchedGraphsForUrl(url.url);
```
JoeCap08055, line 70:
```suggestion
  async getWebhooksForUrl(@Query() { url }: UrlParam): Promise<string[]> {
    return this.apiService.getWatchedGraphsForUrl(url);
```

apps/graph-api/src/controllers/v1/webhooks-v1.controller.ts
```diff
@@ -106,8 +107,8 @@ export class WebhooksControllerV1 {
     example: '2',
     description: 'MSA for which to remove all webhook registrations',
   })
-  deleteWebhooksForMsa(@Param('msaId') msaId: string): Promise<void> {
-    return this.apiService.deleteWebhooksForUser(msaId);
+  deleteWebhooksForMsa(@Param() msaId: MsaIdParam): Promise<void> {
+    return this.apiService.deleteWebhooksForUser(msaId.msaId);
```
JoeCap08055, line 111:
```suggestion
  deleteWebhooksForMsa(@Param() { msaId }: MsaIdParam): Promise<void> {
    return this.apiService.deleteWebhooksForUser(msaId);
```

apps/graph-api/src/controllers/v1/webhooks-v1.controller.ts
```diff
@@ -120,7 +121,7 @@ export class WebhooksControllerV1 {
     example: 'http://localhost/webhook',
     description: 'URL for which to un-watch all MSAs',
   })
-  deleteAllWebhooksForUrl(@Query('url') url: string): Promise<void> {
-    return this.apiService.deleteWebhooksForUrl(url);
+  deleteAllWebhooksForUrl(@Query() url: UrlParam): Promise<void> {
+    return this.apiService.deleteWebhooksForUrl(url.url);
```
JoeCap08055, line 125:
```suggestion
  deleteAllWebhooksForUrl(@Query() { url }: UrlParam): Promise<void> {
    return this.apiService.deleteWebhooksForUrl(url);
```

apps/graph-api/src/guards/read-only-deployment-guard.service.ts
JoeCap08055: We'll refactor this to `libs/utils`, but we can't until after we deal with the `ConfigService` dependency.

libs/types/src/dtos/common/params.dto.ts
JoeCap08055, line 1: nit: rename classes to `xxxDto` to be consistent with the current naming pattern

libs/types/src/dtos/common/params.dto.ts
```diff
@@ -0,0 +1,17 @@
+// eslint-disable-next-line max-classes-per-file
+import { ApiProperty } from '@nestjs/swagger';
+import { IsNotEmpty, IsNumberString, IsUrl } from 'class-validator';
+
+export class MsaIdParam {
+  @ApiProperty({ type: String, description: 'MSA Id of account' })
+  @IsNumberString({ no_symbols: true })
```
JoeCap08055, line 7: I'd actually be in favor of creating a custom validation decorator, that can check not only for a valid numeric integer string, but also validate that it's > 0 (or any other simple validations on an MSA that don't rely on the chain)

libs/types/src/dtos/graph/graph-key-pair.dto.ts
```diff
@@ -2,17 +2,24 @@ import { IsEnum, IsNotEmpty, IsString } from 'class-validator';
...
 import { HexString } from '@polkadot/util/types';
 import { ApiProperty } from '@nestjs/swagger';
+import { IsHexValue } from '#utils/decorators/is-hex-value.decorator';
 
 // DTO for the graph key pair
 export class GraphKeyPairDto {
-  @IsNotEmpty()
-  @IsString()
-  @ApiProperty({ description: 'Public graph encryption key as a hex string (prefixed with "0x")', type: String })
+  @IsHexValue({ minLength: 64, maxLength: 64, message: 'publicKey should be a 32 bytes value in hex format!' })
```
JoeCap08055, line 9: I wonder, is it worth it to write a decorator that's able to validate that a hex string decodes to a valid public key generated by a specific algorithm?

tsconfig.json
JoeCap08055: You'll also need to add this path mapping to `jest.config.json`

mattheworris (approved):
- Read through the code
- Looks good, agree with other comments

🚢 it

tsconfig.json
aramikm: Great catch!

libs/types/src/dtos/common/params.dto.ts
aramikm: I agree. I'll create customer MsaId and schemaId validators in here. 
On the side note `no_symbol` forces the numbers to be > 0

libs/types/src/dtos/graph/graph-key-pair.dto.ts
aramikm: I have an issue for semantics checks and will add this inside that.

JoeCap08055:
Regarding e2e tests: if we have a working e2e test framework, then it's worth while to add tests, because the test scenarios themselves for endpoint validation won't change--only the test framework setup will change, and we'll add additional tests for worker scenarios.

But if you want to add tests in a separate PR, that's fine.

aramikm:
I think it would makes sense to add those tests when the framework is ready simply due to improved velocity of writing and verifying them compared to current state that if something goes wrong I would have to dig to see if my written test is the issue or the framework setup.

PR #516, opened 2024-09-13
content-watcher-api: added validation

### LINKED ISSUE #499: content-watcher-api endpoints data context verification
checking duplicates, incompatible inputs, sanitization,...

aramikm:
# Problem

Adds extra verification on dtos

Close #499

## Steps to Verify:

1. run the api
2. open swagger ui
3. call endpoints with invalid types or amounts or requests


libs/types/src/dtos/content-watcher/subscription.webhook.dto.ts
```diff
@@ -1,23 +1,27 @@
...
   })
   url: string; // Webhook URL
 
   @IsArray()
+  @ArrayNotEmpty()
+  @IsEnum(AnnouncementTypeName, { each: true })
   @ApiProperty({
     description: 'Announcement types to send to the webhook',
-    example: ['Broadcast', 'Reaction', 'Tombstone', 'Reply', 'Update'],
+    example: ['broadcast', 'reaction', 'tombstone', 'reply', 'update'],
```
JoeCap08055, line 24:
```suggestion
    isArray: true,
    example: ['broadcast', 'reaction', 'tombstone', 'reply', 'update'],
    enum: AnnouncementTypeName,
    enumName: 'AnnouncementTypeName',
```

JoeCap08055 (commented):
I'm thinking, it would make our endpoint validations across the board cleaner, more concise & more readable/understandable if we write some custom validators for commonly used input types, such as:
* MSA ID/Provider ID
* Public control key
* Account address
* Public/private encryption keys
* Block number/block hash

mattheworris (approved):
Nice work, 🚢 it.

aramikm:
@JoeCap08055 That is a great idea. I've created custom validators for
- Generic Hex values which can be used for public and private keys

I think as you said we can add more customized ones. The issue is that if I add it in here then I've have to decoflict with other open PRs. I'll add this in the last PR so we can go forward with less conflicts.

PR #518, opened 2024-09-17
Content-publisher api: more validations

### LINKED ISSUE #494: content-publishing-api endpoints data syntax verification
Check type, length, format, ...

aramikm:
# Problem

Closes #494
# Solution
- moved existed decorators into utils
- enabled base32 regex verification for ContentHash
- Added `ApiProperty` definitions on existing DTOs;
- minor bugfixes and refactors

## Steps to Verify:

1. Run the content-publishing-api
2. Open swagger
3. Try submitting wrong requests


apps/content-publishing-api/src/api.config.spec.ts
```diff
@@ -12,6 +12,7 @@ describe('Content Publishing API Config', () => {
     API_PORT: undefined,
     // API_TIMEOUT_MS: undefined,
     FILE_UPLOAD_MAX_SIZE_IN_BYTES: undefined,
+    FILE_UPLOAD_COUNT_LIMIT: undefined,
```
JoeCap08055, line 15: Should also add 2 test cases for `validateMissing` and `shouldFailBadValues` in the 'invalid environment' section, since this is a required config.

libs/utils/src/decorators/is-dsnp-content-uri.decorator.ts
```diff
@@ -0,0 +1,49 @@
...
+  // eslint-disable-next-line func-names
+  return function (object: object, propertyName: string) {
+    registerDecorator({
+      name: 'isDsnpContentURI',
+      target: object.constructor,
+      propertyName,
+      options: validationOptions,
+      validator: {
+        validate(value: unknown, _args: ValidationArguments) {
+          const re = /^(?<protocol>.*):\/\/(?<msaId>[\d]*)\/(?<contentHash>.*)$/;
```
JoeCap08055, line 15: Should we allow for an optional trailing '/'? What does the DSNP spec say?

JoeCap08055: DSNP spec doesn't indicate that it's allowed; let's leave this alone, then.

libs/utils/src/decorators/is-msa-id.decorator.ts
```diff
@@ -1,5 +1,12 @@
 import { registerDecorator, ValidationArguments, ValidationOptions } from 'class-validator';
 
+export function isValidMsa(value: unknown): boolean {
+  // no leading zeros are accepted since this would cause different DsnpUserUri to be valid for the same user
+  const re = /^[1-9][0-9]*$/;
+
+  return (typeof value === 'string' && re.test(value)) || typeof value === 'number';
```
JoeCap08055, line 7: Should probably validate that it doesn't exceed u64... either:
1. Convert to `bigint` and compare to 2^64
2. Add a max length to the RE for the longest-length allowable bigint string (less precise, but should be fine as a sanity check, as we're unlikely to ever get to the limit for a real MSA ID)

libs/utils/src/decorators/is-int-value.decorator.ts
```diff
@@ -19,7 +19,7 @@ export function IsIntValue(validationOptions?: IsIntValueValidationOption) {
 
           if ((typeof value === 'string' && re.test(value)) || typeof value === 'number') {
             const numberValue = BigInt(value);
-            if (validationOptions.minValue !== undefined) {
+            if (validationOptions.maxValue !== undefined) {
```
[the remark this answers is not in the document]
aramikm: there is an already existing IsInt() decorator that support that usecase. Our usecase was more limiting the range of the int mainly on the min side and sometimes on both sides. But I can make both of them optional anyway.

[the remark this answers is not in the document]
JoeCap08055: That might be best; it's confusing to have to use one decorator if you want a `minValue`, and another decorator if you don't.

Also, it might be useful to accept another option, `accept_strings: boolean` that controls whether the value must be an actual number, or if it can also be a string representation of a number. The `class-validator` package has `@IsNumber()` and `@IsNumberString()`, both of which take some options worth examining.

Actually, the more I think about it, in the interest of velocity & small PRs, let's go ahead with this and consider enhancing our stable of decorators in a future PR.

JoeCap08055 (approved):

PR #519, opened 2024-09-17
Chore add validation tests

mattheworris:
# Problem

As developer, I would like to have tests for as many logic paths as possible.

# Solution

Implement tests for the validation logic added to the revokeDelegation endpoint.

## Steps to Verify:

1. Execute `./scripts/restart-chain-docker.sh`
2. Execute `make test-e2e-account`


apps/account-api/src/services/delegation.service.ts
```diff
@@ -47,26 +48,29 @@ export class DelegationService {
...
-      }
+    // Validate that the providerId is a registered provider, also checks for valid MSA ID
+    const providerRegistry = await this.blockchainService.getProviderToRegistryEntry(providerId);
+    if (!providerRegistry) {
+      throw new HttpException('Provider not found', HttpStatus.BAD_REQUEST);
+    }
 
-      // Validate that delegations exist for this msaId
+    // Validate that delegations exist for this msaId
+    try {
```
JoeCap08055, line 66: 2 things here:
1. If `getDelegation()` returns anything at all, it will be a delegation for the current provider. So checking the providerId is unnecessary.
2. A delegation object would still be returned if the delegation was already revoked; you need to check `revokedAt` and compare to the current block number. (Actually, just that it's non-zero is fine; we only care about which block it was revoked at if we're checking for a delegation a some point in time past)

apps/account-api/test/delegations.controller.e2e-spec.ts
```diff
@@ -104,6 +105,52 @@ describe('Delegation Controller', () => {
...
+    await request(httpServer).get(getPath).expect(400).expect({
+      statusCode: 400,
+      message: 'Invalid accountId',
+    });
+  });
+
+  it('(GET) /v1/delegation/revokeDelegation/:accountId/:providerId with valid accountId: no msa', async () => {
+    const providerId = provider.msaId?.toString();
+    const keyring = new Keyring({ type: 'sr25519' });
+    const keypair = keyring.addFromUri('//Bob');
```
JoeCap08055, line 122: Perhaps best to not use `//Bob` here; changes to our chain scenario setup might provision a user for `//Bob` at some point. Since you don't need a token-holding user, you can use a derived URI like `//Bob//invalidKeypair`

apps/account-api/test/delegations.controller.e2e-spec.ts
```diff
@@ -104,6 +105,52 @@ describe('Delegation Controller', () => {
...
+    await request(httpServer).get(getPath).expect(404).expect({
+      statusCode: 404,
+      message: 'MSA ID for account not found',
+    });
+  });
+
+  it('(GET) /v1/delegation/revokeDelegation/:accountId/:providerId with invalid providerId', async () => {
+    const { keypair } = users[1];
+    const accountId = keypair.address;
+    const invalidProviderId = '9999';
```
JoeCap08055, line 134: Maybe use an MSA from the provisioned `users[]` array here; they're guaranteed not to be registered Providers. And if you're looking for a value that's guaranteed not a registered MSA, you can query `api.query.msa.currentMsaIdentifierMaximum()` and add > 1 to it.

apps/account-api/test/delegations.controller.e2e-spec.ts
```diff
@@ -104,6 +105,52 @@ describe('Delegation Controller', () => {
...
+    await request(httpServer).get(getPath).expect(400).expect({
+      statusCode: 400,
+      message: 'Provider not found',
+    });
+  });
+
+  it('(GET) /v1/delegation/revokeDelegation/:accountId/:providerId with no delegations', async () => {
+    const providerId = provider.msaId?.toString();
+    const keyring = new Keyring({ type: 'sr25519' });
+    const keypair = keyring.addFromUri('//Alice');
```
JoeCap08055, line 145: `//Alice` is already our Provider, maybe use a different keypair here? If you want a registered MSA that's guaranteed not to have a current delegation to `//Alice`, you have 2 options:
1. Create a new keypair + MSA in the test
2. Revoke a delegation from one of the existing `users[]` in the test first.

Actually, both 1 & 2 should be separate test scenarios--(1) tests a `null` delegation, while (2) tests a revoked delegation.

apps/account-api/test/delegations.controller.e2e-spec.ts
claireolmstead: look to other tests where we get user[]

apps/account-api/test/delegations.controller.e2e-spec.ts
```diff
@@ -104,6 +105,51 @@ describe('Delegation Controller', () => {
...
+  // it('(GET) /v1/delegation/revokeDelegation/:accountId/:providerId with revoked delegations', async () => {
+  //   const providerId = provider.msaId?.toString();
+  //   // last user has no delegations because they were revoked
+  //   const { keypair } = users[3];
+  //   const getPath: string = `/v1/delegation/revokeDelegation/${keypair.address}/${providerId}`;
+  //   await request(httpServer).get(getPath).expect(404).expect({
+  //     statusCode: 404,
+  //     message: 'Delegation not found',
+  //   });
+  // });
```
claireolmstead, line 151: leftover?

claireolmstead (dismissed):
A couple changes here, but overall looks really good!

apps/account-api/test/delegations.controller.e2e-spec.ts
mattheworris: All `user[]` have an msa. This test needs an `accountId` without an MSA.

package.json
```diff
@@ -73,6 +73,7 @@
     "test": "npm run test:account ; npm run test:content-publishing ; npm run test:content-watcher ; npm run test:graph",
     "test:verbose": "jest --coverage --verbose",
     "test:e2e:account": "dotenvx run -f env-files/account.template.env -- jest --runInBand --detectOpenHandles --testRegex 'account-api/.*\\.e2e-spec\\.ts'",
+    "test:e2e:delegations": "dotenvx run -f env-files/account.template.env -- jest --runInBand --detectOpenHandles --testRegex 'account-api/test/delegations.controller.e2e-spec.ts' -t 'v1/delegation/revokeDelegation'",
```
mattheworris, line 76: A useful example for how to isolate tests for development

apps/account-api/test/delegations.controller.e2e-spec.ts
```diff
@@ -119,6 +120,58 @@ describe('Delegation Controller', () => {
   describe('/v1/delegation/revokeDelegation', () => {
     let delegationObj: RevokeDelegationPayloadResponseDto;
 
+    it('(GET) /v1/delegation/revokeDelegation/:accountId/:providerId with invalid accountId', async () => {
+      const providerId = provider.msaId?.toString();
+      const { keypair } = users[1];
+      const invalidAccountId = `${keypair.address.slice(0, -1)}5`;
```
JoeCap08055, line 126: What if the keypair already ends in '5'?
You could just generate a keypair here, guaranteed not to be an MSA. For instance:
```typescript
const invalidKeypair = new Keyring({ type: 'sr25519' }).createFromUri('//Alice//invalidUser');
```

apps/account-api/test/delegations.controller.e2e-spec.ts
```diff
@@ -119,6 +120,58 @@ describe('Delegation Controller', () => {
...
+      const getPath: string = `/v1/delegation/revokeDelegation/${invalidAccountId}/${providerId}`;
+      await request(httpServer).get(getPath).expect(400).expect({
+        statusCode: 400,
```
JoeCap08055, line 129: nit: use enum instead of numeric code, `HttpStatus.BAD_REQUEST`

apps/account-api/test/delegations.controller.e2e-spec.ts
```diff
@@ -119,6 +120,58 @@ describe('Delegation Controller', () => {
...
+        message: 'Invalid accountId',
+      });
+    });
+
+    it('(GET) /v1/delegation/revokeDelegation/:accountId/:providerId with valid accountId: no msa', async () => {
+      const providerId = provider.msaId?.toString();
+      const keyring = new Keyring({ type: 'sr25519' });
+      const keypair = keyring.addFromUri('//Bob//InvalidMSA');
```
JoeCap08055, line 137:
```suggestion
      const keypair = new Keyring({ type: 'sr25519' }).createFromUri('//Bob//InvalidMSA');
```

JoeCap08055: THere are a couple of tests that use a valid public key address that is not mapped to an MSA; maybe hoist this to the top of the file in a global `nonMsaAccountAddress` variable?

apps/account-api/test/delegations.controller.e2e-spec.ts
```diff
@@ -119,6 +120,58 @@ describe('Delegation Controller', () => {
...
+        message: 'Delegation already revoked',
+      });
+    });
+
+    it('(GET) /v1/delegation/revokeDelegation/:accountId/:providerId with no delegations', async () => {
+      const providerId = provider.msaId?.toString();
+      const getPath: string = `/v1/delegation/revokeDelegation/${undelegatedUser.keypair.address}/${providerId}`;
+      await request(httpServer).get(getPath).expect(404).expect({
+        statusCode: 404,
+        message: 'Delegation not found: Error: Failed to find the delegation.',
```
JoeCap08055, line 171:
```suggestion
        message: 'Delegation not found.',
```

apps/account-api/test/delegations.controller.e2e-spec.ts
mattheworris: The point of this test is an invalid accountId, so a new keypair doesn't hit the code path.
Point taken about the 5, I added `5H` instead so it passes dto validation, but not `isValidAddress`

mattheworris: `isValidAddress` can probably be moved to the DTO validation with decorators, but leaving it here for now. Also, there was a version that had an `ss58` validation decorator for the accountId, but I took it out, as the validation refactor is going to happen for v1.2.

JoeCap08055 (approved):
All my comments addressed, pipeline is green. 🚢 it!

PR #521, opened 2024-09-18
feat: add new 'v2' /delegations endpoints and e2e tests

JoeCap08055:
- Create a `v2` controller `/delegations`
- Change API path from `/delegation` -> `/delegations` to better conform to our REST nomenclature (also reflecting the fact that we can now fetch multiple)
- Add `GET /v2/delegations/:msaId` to get all delegations for an MSA
- Add `GET /v2/delegations/:msaId/:providerId` to get an MSA's delegations to a specific provider
- Use NestJS API versioning to add the `/v1/` and `/v2/` path components instead of hard-coding
- Change response type to not use PolkadotJS SCALE data types; instead create our own DTO so that the generated OpenAPI interface is complete (`/v1` endpoint with PolkadotJS response type was opaque for the response type)
- Update chain setup to add un-delegated and revoked MSAs
- Add E2E tests for new endpoints

# Problem

problem statement, including
Link to GitHub Issue(s):

# Solution

What I/we did to solve this problem

with @pairperson1

## Steps to Verify:

1. What to do next
1. Any other instructions
1. Expected behavior
1. Suggestions for testing

## Screenshots (optional):


jest.config.json
```diff
@@ -1,5 +1,6 @@
 {
   "verbose": true,
+  "silent": true,
```
mattheworris, line 3: Do `verbose` and `silent` affect different output? Seems like they would be opposites.

JoeCap08055: `silent` suppresses console output from the code under test
`verbose` enables detailed test pass/fail reporting output

apps/account-api/src/controllers/v1/delegation-v1.controller.ts
```diff
@@ -20,8 +20,8 @@ import {
 } from '@nestjs/common';
 import { ApiBody, ApiCreatedResponse, ApiOkResponse, ApiOperation, ApiTags } from '@nestjs/swagger';
 
-@Controller('v1/delegation')
-@ApiTags('v1/delegation')
+@Controller({ version: '1', path: 'delegation' })
```
claireolmstead, line 23: why is this version a string and the version in the delegation-v2.controller.ts an array?

mattheworris (approved):
- Read through changes, looks good.

:shipit:

apps/account-api/test/delegations.controller.e2e-spec.ts
```diff
@@ -20,20 +22,25 @@ let publicFollowsSchema: Schema | undefined;
...
+let invalidMsaId: string;
+let msaNonProviderId: string;
 
 describe('Delegation Controller', () => {
   let app: INestApplication;
   let module: TestingModule;
 
   beforeAll(async () => {
-    ({ maxMsaId, provider, users } = await setupProviderAndUsers());
+    ({ maxMsaId, provider, revokedUsers: revokedUser, undelegatedUser, users } = await setupProviderAndUsers());
```
claireolmstead, line 33: a little confusing to have a plural -> singular. If a value is plural, I would assume it's an array.

jest.config.json
claireolmstead, line 3: these two are conflicting. what does it do in the config to define them both?

libs/account-lib/src/blockchain/blockchain.service.ts
```diff
@@ -401,8 +403,20 @@ export class BlockchainService implements OnApplicationBootstrap, OnApplicationS
     providerId: AnyNumber,
   ): Promise<CommonPrimitivesMsaDelegation | null> {
     const delegationResponse = await this.api.query.msa.delegatorAndProviderToDelegation(msaId, providerId);
-    if (delegationResponse.isSome) return delegationResponse.unwrap();
-    return null;
+    return delegationResponse.unwrapOr(null);
```
claireolmstead, line 406: this is cool!

apps/account-api/src/controllers/v1/delegation-v1.controller.ts
JoeCap08055: https://github.com/ProjectLibertyLabs/gateway/pull/513/files#r1765252182

claireolmstead (approved):
LGTM. A couple of questions

jest.config.json
JoeCap08055: https://github.com/ProjectLibertyLabs/gateway/pull/521#discussion_r1765272223

apps/account-api/test/delegations.controller.e2e-spec.ts
JoeCap08055: Good catch; that was a VSCode refactor anomaly. Should be singular; I'll fix.

PR #554, opened 2024-09-23
Refactor content-watcher service configs

### LINKED ISSUE #547: Refactor `content-watcher-service` configs

JoeCap08055:
This PR refactors `ConfigModule` and `ConfigService` for `account-service`.

- Eliminate custom config service class and rely simply on built-in NestJS `ConfigModule`, using namespaced configs for individual modules
- Move config property configuration into module-specific config, ie 'blockchain', 'cache', 'queue', 'account-api', 'account-worker'. (NOTE, as each module is refactored in the future, these module configs will be merged with the corresponding module configs from the other services, but for now they will still be duplicated across the services)
- Update & add unit tests to cover new config setup

Closes #547


libs/content-watcher-lib/src/pubsub/pubsub.config.ts
```diff
@@ -0,0 +1,23 @@
...
+
+export interface IPubSubConfig {
+  webhookMaxRetries: number;
+  webhookRetryIntervalSeconds: number;
+}
+
+export default registerAs('pubsub', (): IPubSubConfig => {
+  const configs: JoiUtils.JoiConfig<IPubSubConfig> = {
+    webhookMaxRetries: {
+      value: process.env.WEBHOOK_FAILURE_THRESHOLD,
```
aramikm, line 13: I'm guessing `WEBHOOK_FAILURE_THRESHOLD` and `webhookMaxRetries` are kinda the same? Having **threshold** in here feels confusing if they are the same.

aramikm (approved):
Looks good!

libs/content-watcher-lib/src/pubsub/pubsub.config.ts
JoeCap08055: Yes. Some of the webhook/health check-related configs are used differently across the different Gateway services; that can be the subject of a whole separate PR.

PR #556, opened 2024-09-24
fix: URLs in configs must be strings, not JS URL objects

JoeCap08055:
# Problem

The recent refactor to configs changed the internal representation of URL config values from string -> JS `URL` object. This caused some problems.

# Solution

Revert to string representation of URLs.


mattheworris (approved):
🚢 it

PR #563, opened 2024-09-26
Account api context checks

### LINKED ISSUE #493: account-api endpoints data context verification
- checking duplicates, incompatible inputs, sanitization,...

aramikm:
# Problem

closes #493 

# Changes
- Added extra verification fro signatures and accountIds
- Checks provided signatures in some endpoints
- removed wrapped `<Bytes></Bytes>` in endpoints that made sense.

## Steps to Verify:
1. run account-api
2. open swagger
3. try incorrect signatures or payloads


apps/account-api/src/controllers/v1/handles-v1.controller.ts
```diff
@@ -103,7 +116,7 @@ export class HandlesControllerV1 {
     try {
       const expiration = await this.handlesService.getExpiration();
       const payload = { baseHandle: newHandle, expiration };
-      const encodedPayload = u8aToHex(u8aWrapBytes(this.handlesService.encodePayload(payload).toU8a()));
+      const encodedPayload = u8aToHex(this.handlesService.encodePayload(payload).toU8a());
```
JoeCap08055, line 119: My only questions here are:
1. Was this working before, and if so, why? Or is is just that we can now simplify the client signing code on the other side?
2. Is this confirmed working as expected now after this change?

apps/account-api/src/services/delegation.service.ts
```diff
@@ -47,36 +46,31 @@ export class DelegationService {
...
 
-    if (providerId) {
-      const isValidProviderId = await this.blockchainService.isValidMsaId(providerId);
-      if (!isValidProviderId) {
-        throw new HttpException('Invalid provider', HttpStatus.BAD_REQUEST);
-      }
-
-      const providerInfo = await this.blockchainService.getProviderToRegistryEntry(providerId);
+    if (providerMsaId) {
+      const providerInfo = await this.blockchainService.getProviderToRegistryEntry(providerMsaId);
```
JoeCap08055, line 59: Are we sure that this is the change we want? Previously, we threw 2 different errors:
* `BAD_REQUEST('Invalid provider')` if the ID supplied for the Provider is not a valid MSA
* `BAD_REQUEST('Supplied ID not a Provider')` if the ID supplied IS an MSA, but IS NOT a registered Provider

I think having 2 different error strings returned is useful to a client in understanding _why_ their request failed.

apps/account-api/src/services/keys.service.ts
```diff
@@ -71,9 +74,7 @@ export class KeysService {
         ],
       };
 
-      const encodedPayload = u8aToHex(
-        u8aWrapBytes(this.blockchainService.createItemizedSignaturePayloadV2Type(payload).toU8a()),
-      );
+      const encodedPayload = u8aToHex(this.blockchainService.createItemizedSignaturePayloadV2Type(payload).toU8a());
```
JoeCap08055, line 77: This is not an extrinsic; it's a payload to be included in an extrinsic call. So don't we want to byte-wrap?

apps/account-api/src/controllers/v1/handles-v1.controller.ts
aramikm: My understanding is that it depended on how the payload was signed.
1- If the payload was signed directly with the keypair then it works as expected since there is no mandetory wrapping for that
2- If the payload was signed by the polkadot wallet as raw data then the wallet would wrap it in Bytes regardless. So Having it already wrapped would cause issues.

I think to b e able to support both cases we probably would like to go the unwrapped route since in option 1 we can always wrap it in bytes before signing.

aramikm: @mattheworris Anything that I might be missing?

libs/account-lib/src/utils/utility.ts
```diff
@@ -18,3 +15,34 @@ export const getSignerForRawSignature = (result: SignerResult): Signer => ({
...
+      hexToU8a(payloadHex),
+      hexToU8a(signatureHex),
+      isHex(accountAddressOrPublicKey) ? hexToU8a(accountAddressOrPublicKey) : accountAddressOrPublicKey,
+    );
+    return {
+      isValid,
+      isWrapped,
+    };
+  } catch (error) {
+    console.error('verifySignature:', error);
```
JoeCap08055, line 42: Either remove or use a `Logger` object.

apps/account-api/src/services/delegation.service.ts
aramikm: I might be missing something but I think we only got rid of the first one not the second

apps/account-api/src/services/keys.service.ts
aramikm: It is the same argument I mentioned before. It depends how it is going to be signed.

To get rid of confusion I have in mind to return different types for raw and wrapped payloads to the consumer would know this is wrapped or not for ease of use.

apps/account-api/src/services/delegation.service.ts
```diff
@@ -47,36 +46,31 @@ export class DelegationService {
...
+    accountId: string,
+    providerMsaId: string,
+  ): Promise<RevokeDelegationPayloadResponseDto> {
     const msaId = await this.blockchainService.publicKeyToMsaId(accountId);
     if (!msaId) {
       throw new HttpException('MSA ID for account not found', HttpStatus.NOT_FOUND);
     }
 
-    if (providerId) {
-      const isValidProviderId = await this.blockchainService.isValidMsaId(providerId);
```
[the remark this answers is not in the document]
JoeCap08055: Validated via decorator semantically, but not on-chain. `BlockchainService.isValidMsaId(msaId)` validates that the given identifier is actually an MSA registered on-chain. (more specifically, validates that `msaId <= api.query.msa.currentMsaIdentifierMaximum()`, so technically it doesn't check if the MSA is retired, but still...)

apps/account-api/src/services/delegation.service.ts
JoeCap08055: Right; we've consolidated 2 errors -> 1 error. Do we want to do that?

apps/account-api/src/services/delegation.service.ts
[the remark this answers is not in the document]
aramikm: Ah I see what you are saying. I think it's a bit overkill to call the chain twice since we already doing it in `getProviderToRegistryEntry` but I don't have any strong opinions and will add that back

apps/account-api/src/controllers/v1/handles-v1.controller.ts
mattheworris: Suggestion: can we add e2e tests for these payloads?

JoeCap08055: We're going to have to do a whole series of tickets for E2E tests, after the initial round of hardening tickets are done, I think.

JoeCap08055 (approved):

libs/types/src/content-announcement/types.gen.ts
```diff
@@ -7,7 +7,6 @@ export enum AnnouncementType {
   Reaction = 4,
   Profile = 5,
   Update = 6,
-  PublicFollows = 113,
```
mattheworris, line 10: Has this moved to a different announcement type? Flagging because SAT uses public follows.

aramikm: this is auto generated. I'm not sure. @JoeCap08055 do you have any idea?

Issue #524, opened 2024-09-19
Implement Authentication Session URL Retrieval in Gateway


p5150j:
## Summary:
As a developer, I want to implement the functionality in the Gateway to retrieve an `AuthenticationSessionUrl,` so that the app backend can initiate the sign-in process.

### Acceptance Criteria:

- Implement the `getAuthenticationSessionUrl()` method in the Gateway.
- Generate and sign the request payload (`SiwaRequest`).
- Return the `AuthenticationSessionUrl ` to the app backend.

Tasks:

1. Develop the `getAuthenticationSessionUrl()` logic in the Gateway.
2. Generate and sign the `SiwaRequest `payload.
3. Ensure proper error handling and logging.
4. Write unit tests for the method.

PR #564, opened 2024-09-27
SIWF V2 GET /siwf: Get the Redirect URL

wilwade:
# Problem

A provider backend using SIWF V2 (also known as SIWA) wants to be able to generate Authentication Redirect URLs with dynamic callback URLs instead of using  a static or pre-generated `signedPayload`.

Closes: #524

# Solution

- Update `SIWF_URL` (aka v1) to the new v1 path
- Added new `SIWF_V2_URL` optional environment variable (defaults to Frequency Access for now)
- Added new V2 accounts controller and e2e test
- Added `WalletV2RedirectRequestDto` with decorators and transformations to validate the inputs
- Added Tests for new (complex) decorators.
- Added new `SiwfV2Service` to handle the processing of the SIWF V2 related tasks
- Updated incorrect docs
- Fixed issue with failing keys.controller test due to mismatched versions of Polkadotjs API.

## Steps to Verify:

1. Run the e2e tests
2. Test it via http://localhost:3000/docs/swagger#/v2%2Faccounts/AccountsControllerV2_getRedirectUrl

## Screenshots (optional):

image


apps/account-api/src/api.module.ts
```diff
@@ -8,13 +8,14 @@ import { AccountQueues as QueueConstants } from '#types/constants/queue.constant
 import { CacheModule } from '#cache/cache.module';
 import {
   AccountsControllerV1,
+  AccountsControllerV2,
   DelegationControllerV1,
   HandlesControllerV1,
   KeysControllerV1,
+  DelegationsControllerV2,
```
wilwade, line 15: Updated this to follow the pattern from the other ones.

apps/account-api/src/controllers/v2/accounts-v2.controller.ts
```diff
@@ -0,0 +1,67 @@
+import { SiwfV2Service } from '#account-api/services/siwfV2.service';
+import { SCHEMA_NAME_TO_ID } from '#types/constants/schemas';
+import { WalletV2RedirectRequestDto } from '#types/dtos/account/wallet.v2.redirect.request.dto';
+import { WalletV2RedirectResponseDto } from '#types/dtos/account/wallet.v2.redirect.response.dto';
+import { Controller, HttpCode, HttpStatus, Logger, HttpException, Query, Get } from '@nestjs/common';
+import { ApiOkResponse, ApiOperation, ApiTags } from '@nestjs/swagger';
+
+// # SIWF Wallet API V2
```
wilwade, line 8: This documentation is a copy of the docs at https://projectlibertylabs.github.io/siwf/v2/docs/ for easy reference while this is being built. Better docs coming once the entire controller is complete

apps/account-api/src/main.ts
```diff
@@ -57,13 +57,15 @@ async function bootstrap() {
       new ValidationPipe({
         whitelist: true,
         transform: true,
+        enableDebugMessages: !!process.env.DEBUG,
```
wilwade, line 60: I can remove this, but it was very helpful for me to see the detailed logs

apps/account-api/src/services/accounts.service.ts
```diff
@@ -1,5 +1,5 @@
 import { Inject, Injectable, Logger } from '@nestjs/common';
-import { validateSignin, validateSignup } from '@projectlibertylabs/siwf';
+import { validateSignin, validateSignup } from '@projectlibertylabs/siwfv1';
```
wilwade, line 2: This is thanks to node aliasing. Both v1 and v2 are installed.

apps/account-api/test/keys.controller.e2e-spec.ts
wilwade, line 1: Fixing for linting

developer-docs/account/ENVIRONMENT.md
```diff
@@ -3,25 +3,27 @@
 ℹ️ Feel free to adjust your environment variables to taste.
 This application recognizes the following environment variables:
 
-| Name                                      | Description                                                                                                                                                                       |                                                Range/Type                                                | Required? | Default |
```
wilwade, line 6: Most lines are formatting for the change in the width

developer-docs/account/ENVIRONMENT.md
```diff
@@ -3,25 +3,27 @@
...
+| `API_BODY_JSON_LIMIT`                     | Api json body size limit in string (some examples: 100kb or 5mb or etc)                                                                                                           |                                                  string                                                  |           |                       1mb                       |
+| `SIWF_URL`                                | SIWF v1: URL for Sign In With Frequency V1 UI                                                                                                                                     |                                                   URL                                                    |           | https://ProjectLibertyLabs.github.io/siwf/v1/ui |
+| `SIWF_V2_URL`                             | SIWF v2: URL for Sign In With Frequency V2 Redirect URL                                                                                                                           |                                                   URL                                                    |           | https://ProjectLibertyLabs.github.io/siwf/v1/ui |
```
wilwade, line 29: Added SIWF_URL as it was missing and new SWIF_V2_URL

developer-docs/account/README.md
wilwade, line 1: Fixing incorrect steps

docker-compose-e2e.account.yaml
```diff
@@ -11,8 +11,7 @@ x-common-environment: &common-environment
   HEALTH_CHECK_MAX_RETRY_INTERVAL_SECONDS: 10
   HEALTH_CHECK_SUCCESS_THRESHOLD: 10
   CAPACITY_LIMIT: '{"type":"percentage", "value":80}'
-  SIWF_URL: 'https://projectlibertylabs.github.io/siwf/ui'
-  SIWF_DOMAIN: 'localhost'
```
wilwade, line 15: Removed the SIWF_DOMAIN as it is not used anywhere

docker-compose-e2e.account.yaml
```diff
@@ -11,8 +11,7 @@ x-common-environment: &common-environment
...
+  SIWF_URL: 'https://projectlibertylabs.github.io/siwf/v1/ui'
```
wilwade, line 14: Updating to the new default, old one still works

libs/account-lib/src/blockchain/blockchain.service.ts
```diff
@@ -648,4 +648,15 @@ export class BlockchainService implements OnApplicationBootstrap, OnApplicationS
   public decodeTransaction(encodedExtrinsic: string) {
     return this.api.tx(encodedExtrinsic);
   }
+
+  public getNetworkType(): 'mainnet' | 'testnet-paseo' | 'unknown' {
```
wilwade, line 652: I went with a typed string instead of enum, but open to changing.

developer-docs/account/README.md
```diff
@@ -183,10 +183,10 @@ In order to run this project you need:
 
 Modify any environment variables in the `.env` file as needed. The complete set of environment variables is documented [here](./ENVIRONMENT.md), and a sample environment file is provided [here](../../env-files/account.template.env).
 
-1. Copy the template values into the .env files.
+1. Copy the template values into the .env.account file.
```
wilwade, line 186: Various fixes based on my walkthrough of it

libs/utils/src/decorators/is-swif-v2-callback-url.decorator.ts
```diff
@@ -0,0 +1,27 @@
+import { registerDecorator, ValidationOptions, ValidationArguments } from 'class-validator';
+
+export function IsSwifV2CallbackUrl(validationOptions?: ValidationOptions) {
+  // eslint-disable-next-line func-names
+  return function (object: Object, propertyName: string) {
+    registerDecorator({
+      name: 'IsSwifV2CallbackUrl',
```
wilwade, line 7: Couldn't come up with a better generic name, so just went with the specific use case.

apps/account-api/src/controllers/v2/accounts-v2.controller.ts
JoeCap08055, line 1: question for discussion:

Regarding API management/versioning, when we create a new version of a controller, should we include unmodified but still-current endpoints? ie, should the entire controller be versioned?

Pros:
- Clients can use just one version of the controller, either all 'v1' or all 'v2', no need to mix-n-match controller versions for different endpoints

Cons:
- For endpoints that didn't change, it's not really a new version, so it's a bit confusing to the client
- If the unchanged endpoints end up changing later, now we need 'v3'

Thoughts?

apps/account-api/src/services/siwfV2.service.spec.ts
```diff
@@ -0,0 +1,137 @@
...
+  it('should be defined', () => {
+    expect(siwfV2Service).toBeDefined();
+  });
+
+  describe('getRedirectUrl', () => {
+    let result: WalletV2RedirectResponseDto;
+
+    beforeAll(async () => {
+      result = await siwfV2Service.getRedirectUrl(exampleCallback, examplePermissions, exampleCredentials);
+    });
```
JoeCap08055, line 64: Since this function has the potential to throw, it would be good to test that it doesn't before proceeding.
```suggestion
    it('Should not throw', async () => {
      expect.assertions(1);
      result = await siwfV2Service.getRedirectUrl(exampleCallback, examplePermissions, exampleCredentials);
      expect(result).toBeDefined();
    });
```

Jest runs all tests in the same `describe` block serially (and awaits any async tests).

libs/utils/src/decorators/ensure-array.decorator.ts
```diff
@@ -0,0 +1,6 @@
+import { Transform } from 'class-transformer';
+
+export const EnsureArray = Transform(({ value }) => {
```
aramikm, line 3: how is this different from `@IsArray()`?

apps/account-api/src/controllers/v2/accounts-v2.controller.ts
wilwade: Hmm... Thinking of this from a development perspective. I expect to "break" a few other endpoints in the account controller. So at least including those doesn't make sense. I think it is fine to have a mix and match. Effectively the endpoints are versioned not the controller?

apps/account-api/src/services/siwfV2.service.spec.ts
```diff
@@ -0,0 +1,137 @@
...
+        imports: [
+          ConfigModule.forRoot({
+            load: [apiConfig, blockchainConfig],
+          }),
+        ],
+        providers: [SiwfV2Service, mockBlockchainServiceProvider],
+      }).compile();
+
+      siwfV2ServiceEndpointTest = module.get<SiwfV2Service>(SiwfV2Service);
+    });
```
JoeCap08055, line 123: All this duplicate test harness setup is overkill, just to control the config environment. Take a look at `testlib/utils.config-tests.ts`. `GenerateMockConfigProvider` can be used to mock the config & enable using `jest.spyon(mockConfigProvider, 'configValue', 'get').mockReturnValue(...)` in specific tests.

libs/utils/src/decorators/is-schema-name.decorator.ts
```diff
@@ -0,0 +1,22 @@
...
+  // eslint-disable-next-line func-names
+  return function (object: Object, propertyName: string) {
+    registerDecorator({
+      name: 'IsSchemaName',
+      target: object.constructor,
+      propertyName,
+      options: validationOptions,
+      validator: {
+        validate(value: any) {
+          return SCHEMA_NAME_TO_ID.has(value);
```
aramikm, line 14: nit: we could sanitize the value like trim it or make it all lower case. This should probably happen during transformation if it's easy to do.

aramikm (approved):
Overall looks good to me some nits and suggestions.

libs/types/src/constants/schemas.ts
```diff
@@ -0,0 +1,25 @@
...
+  ['dsnp.broadcast@v2', 17],
+  ['dsnp.dsnp-content-attribute-set@v1', 12],
+  ['dsnp.ext-content-attribute-set@v1', 13],
+  ['dsnp.private-connections', 10],
+  ['dsnp.private-follows', 9],
+  ['dsnp.profile-resources@v1', 15],
+  ['dsnp.profile@v1', 6],
+  ['dsnp.public-follows', 8],
+  ['dsnp.public-key-assertion-method@v1', 14],
+  ['dsnp.public-key-key-agreement', 7],
```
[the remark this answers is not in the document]
wilwade: That was a typo. Fixing...

apps/account-api/src/services/siwfV2.service.ts
```diff
@@ -0,0 +1,94 @@
...
+        anyOf: [VerifiedEmailAddressCredential, VerifiedPhoneNumberCredential],
+      });
+    } else if (requestCredentials.includes('VerifiedEmailAddressCredential')) {
+      credentials.push({
+        anyOf: [VerifiedEmailAddressCredential],
+      });
+    } else if (requestCredentials.includes('VerifiedPhoneNumberCredential')) {
+      credentials.push({
+        anyOf: [VerifiedPhoneNumberCredential],
+      });
```
JoeCap08055, line 43: Why do we need `anyOf` for the case of just one of these?

apps/account-api/src/services/siwfV2.service.ts
```diff
@@ -0,0 +1,94 @@
...
+    if (requestCredentials.includes('VerifiedGraphKeyCredential')) {
+      credentials.push(VerifiedGraphKeyCredential);
+    }
+    return credentials;
+  }
+
+  // Default to the default "production" and "staging" endpoints for mainnet and testnet-paseo
+  private swifV2Endpoint(): string {
+    const { siwfV2Url }: IAccountApiConfig = this.apiConf;
+    if (siwfV2Url !== '') return siwfV2Url;
```
JoeCap08055, line 55: This may never match--you've defined a Joi validation rule for this config of `Joi.empty('')`, meaning that an empty string will be converted to `undefined`.

apps/account-api/test/e2e-setup.mock.spec.ts
```diff
@@ -47,7 +47,7 @@ export async function removeExtraKeysFromMsa({
   msaId,
   keypair,
 }: {
-  msaId?: MessageSourceId | undefined;
+  msaId?: MessageSourceId | string | undefined;
```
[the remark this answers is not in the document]
JoeCap08055: Might as well use `AnyNumber`, then

apps/account-api/src/services/siwfV2.service.spec.ts
wilwade: A thrown error on a jest test beforeAll still counts as a failure. Do we think that isn't good enough?

```
 FAIL  apps/account-api/src/services/siwfV2.service.spec.ts (5.296 s)
  ● SiwfV2Service › getRedirectUrl › Should return a valid signedRequest

    Failed to get SIWF V2 Redirect URL
```

libs/account-lib/src/blockchain/blockchain.service.ts
JoeCap08055: I'm fine either way, but would prefer to see it exported as a reusable type:
```
export type NetworkType = 'mainnet' | 'testnet-paseo' | 'unknown';
```

libs/types/src/constants/schemas.ts
JoeCap08055, line 1: This seems fine for now, since dev/testnet/mainnet will all have matching schema IDs for `dsnp` and `frequency`-namespaced schemas, at least.

In the future we should probably publish this in a consumable package.

libs/types/src/dtos/account/wallet.v2.redirect.request.dto.ts
JoeCap08055, line 1: Unless the property is an enum, with my pending PR you no longer need to use `@ApiProperty()`. You can just use JSDoc comments for the description & example, and use standard `class-validator` decorators, and `?` annotation for optional properties.

libs/utils/src/decorators/ensure-array.decorator.ts
wilwade: Good question. This is a transformer, so it ensures that it is an array even if it comes in as not an array. GET Query strings will default to a string not an array if you only send one value, this will transform that single value into an array of 1.

Will add some docs to that effect.

libs/utils/src/decorators/is-schema-name.decorator.ts
wilwade: I'm ok for now with this being specific, but we should look at that for several places.

apps/account-api/src/services/siwfV2.service.ts
wilwade: Good point. In theory either should work. I'll need to check with Peter before changing it, so that will likely be in another PR.

apps/account-api/src/services/siwfV2.service.ts
wilwade: Ah good point. Updating to:
```suggestion
    if (siwfV2Url) return siwfV2Url;
```

libs/types/src/constants/schemas.ts
wilwade: I think eventually it moves to fetching from the chain is more likely, but either way, I agree we have too many of these lists already.

libs/types/src/dtos/account/wallet.v2.redirect.request.dto.ts
wilwade: 👍 I can update after your pr either in this one, or in a follow up.

libs/types/src/dtos/account/wallet.v2.redirect.request.dto.ts
```diff
@@ -0,0 +1,38 @@
...
+import { IsSchemaName } from '#utils/decorators/is-schema-name.decorator';
+import { IsSwifV2CallbackUrl } from '#utils/decorators/is-swif-v2-callback-url.decorator';
+import { ApiProperty } from '@nestjs/swagger';
+import { IsOptional } from 'class-validator';
+
+export class WalletV2RedirectRequestDto {
+  @ApiProperty({
+    description: 'The URL that will be called when the authentication process is completed.',
+    required: true,
+    type: String,
```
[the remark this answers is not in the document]
wilwade: Added some this morning as I realized I had them all ready built thanks to the siwf docs

libs/types/src/dtos/account/wallet.v2.redirect.request.dto.ts
JoeCap08055: Let's leave this, and I'll update in my PR, since I'll have to merge/rebase anyway after this is merged, and mine's still under review.

JoeCap08055 (approved):
Looks good!

PR #569, opened 2024-09-30
Chore/refactor openapi generation

### LINKED ISSUE #568: Refactor OpenApi/Swagger spec generation to a common module

JoeCap08055:
This PR does the following:

- refactors the code to generate OpenAPI specs from the NestJS controllers into a common module that can be shared between services.
- removes the path that automatically re-generated the OpenAPI spec file every time the service was launched.
- adds the `@nestjs/swagger` CLI plugin. This eliminates the need to manually generate metadata, so that has been removed from all scripts
- Eliminate manual use of `@ApiProperty()` decorators in _most_ places, and rely instead on the `@nestjs/swagger` package to infer endpoint types from the metadata. A notable exception to this is cases where properties are enums.
- Some type cleanup to make sure the generated OpenAPI spec matches the actual types being returned as fully as possible 
- Added some cleanup to the `CacheMonitorService`; found that registered listeners on the Redis connections were causing application termination to hang.

Closes #568


.github/workflows/common/openapi-build/action.yaml
```diff
@@ -22,10 +22,6 @@ runs:
       run: npm ci
       shell: bash
 
-    - name: Generate Swagger Metadata
-      run: npm run generate:metadata:${{ inputs.service }}
-      shell: bash
-
```
JoeCap08055, line 28: We no longer need to manually generate metadata

apps/account-api/src/build-openapi.ts
JoeCap08055, line 1: This logic has been refactored, part to a shared component, part to `main.ts` for each API service.

apps/account-api/src/main.ts
```diff
@@ -42,6 +42,18 @@ async function bootstrap() {
...
+    description: 'Account Service API',
+    version: '1.0',
+  });
+
+  const args = process.argv.slice(2);
+  if (args.find((v) => v === '--writeOpenApi')) {
+    writeOpenApiFile(swaggerDoc, './openapi-specs/account.openapi.json');
+    process.exit(0);
+  }
+
```
JoeCap08055, line 56: Moved here from `buld-openapi.ts`. Common to both runtime Swagger UI, and OpenAPI spec file generation.

apps/account-api/src/swagger_config.ts
JoeCap08055, line 1: Refactored to shared component (same for this file across all services)

libs/openapi/src/openapi.ts
JoeCap08055, line 1: This was refactored from each service's `swagger_config.ts`, and made so that service-specific parameters get passed in from `main.ts`

aramikm (commented):
Added some nits and questions. There are many things that needed to be verified since this can have effect of other things such as whitelisting checks and ...

- we might be able to remove `metadata.raw.ts.template`

libs/types/src/dtos/account/graphs.request.dto.ts
```diff
@@ -18,51 +18,49 @@ export enum ItemActionType {
 }
 
 export class ItemActionDto {
-  @ApiProperty({
```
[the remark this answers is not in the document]
JoeCap08055: Hmm, yes... that discovery came later in my refactor efforts; I thought I went back and got them all. I'll fix.

libs/types/src/dtos/account/accounts.response.dto.ts
```diff
@@ -1,59 +1,49 @@
 /* eslint-disable max-classes-per-file */
-import { ApiProperty, ApiPropertyOptional } from '@nestjs/swagger';
 import { IsNotEmpty, IsOptional } from 'class-validator';
 import { HexString } from '@polkadot/util/types';
 import { IsHexValue } from '#utils/decorators';
 import { IsAccountIdOrAddress } from '#utils/decorators/is-account-id-address.decorator';
 
 export class HandleResponseDto {
-  @ApiProperty()
```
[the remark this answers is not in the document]
JoeCap08055: I'll test--but I don't think the `@ApiProperty` decorator counts as far as this is concerned; it's only for metadata, not any kind of payload validation.

However, the reason there are no other decorators is because this is a *response* object only--the ONLY reason to make it a concrete DTO class rather than a simple interface or typedef is so that the Swagger/OpenAPI generation will work on it to document the response payload properly.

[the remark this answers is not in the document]
JoeCap08055: 👇🏻 same with the other classes in this file

apps/graph-api/src/main.ts
```diff
@@ -53,7 +65,6 @@ async function bootstrap() {
     app.useGlobalInterceptors(new TimeoutInterceptor(apiConf.apiTimeoutMs));
     app.useBodyParser('json', { limit: apiConf.apiBodyJsonLimit });
```
JoeCap08055, line 67:
```suggestion
    app.useBodyParser('json', { limit: apiConf.apiBodyJsonLimit });

    initializeSwaggerUI(app, swaggerDoc);
```

apps/content-watcher/src/main.ts
```diff
@@ -51,7 +64,6 @@ async function bootstrap() {
     app.useGlobalInterceptors(new TimeoutInterceptor(config.apiTimeoutMs));
     app.useBodyParser('json', { limit: config.apiBodyJsonLimit });
```
JoeCap08055, line 66:
```suggestion
    app.useBodyParser('json', { limit: config.apiBodyJsonLimit });

    initializeSwaggerUI(app, swaggerDoc);
```

apps/content-publishing-api/src/main.ts
```diff
@@ -53,7 +66,6 @@ async function bootstrap() {
     );
     app.useGlobalInterceptors(new TimeoutInterceptor(config.apiTimeoutMs));
     app.useBodyParser('json', { limit: config.apiBodyJsonLimit });
```
JoeCap08055, line 68:
```suggestion
    app.useBodyParser('json', { limit: config.apiBodyJsonLimit });
    
    initializeSwaggerUI(app, swaggerDoc);
```

apps/content-publishing-api/src/controllers/v1/profile.controller.v1.ts
```diff
@@ -17,7 +17,7 @@ export class ProfileControllerV1 {
   @Put(':msaId')
   @ApiOperation({ summary: "Update a user's Profile" })
   @HttpCode(202)
-  @ApiResponse({ status: '2XX', type: AnnouncementResponseDto })
+  // @ApiResponse({ status: '2XX', type: AnnouncementResponseDto })
```
JoeCap08055, line 20:
```suggestion
```

libs/cache/src/cache-monitor.service.ts
```diff
@@ -7,19 +7,33 @@ import { MILLISECONDS_PER_SECOND } from 'time-constants';
...
+    if (this.timeout) {
+      clearTimeout(this.timeout);
+      this.timeout = null;
+    }
+
+    this.clients.forEach((client) => {
+      client.removeAllListeners();
+    });
+  }
+
```
JoeCap08055, line 29: Added this cleanup--makes the application exit more cleanly instead of hanging.

aramikm (approved):
Ship it

PR #588, opened 2024-10-03
refactored ipfs

### LINKED ISSUE #587: Refactor IPFS service
### Context

IPFS service is duplicate across multiple places and we need to clean it up

aramikm:
# Problem

closes #587 

# Changes
- refactored ipfs related services and tools into `libs/storage/ipfs`
- we can extend the storage with AWS S3 or other persistent layers in future (even though that would require writing a storage strategy selector).


apps/content-publishing-api/src/controllers/v1/development.controller.v1.ts
```diff
@@ -64,20 +62,6 @@ export class DevelopmentControllerV1 {
     return job;
   }
 
-  @Get('/asset/:assetId')
```
JoeCap08055, line 67: Any particular reason we don't want to keep this endpoint for dev/debug?

apps/content-publishing-worker/src/worker.module.ts
```diff
@@ -58,6 +59,7 @@ import { ContentPublishingQueues as QueueConstants } from '#types/constants';
       ignoreErrors: false,
     }),
     ScheduleModule.forRoot(),
+    IPFSStorageModule,
```
JoeCap08055, line 62: This probably doesn't need to be imported here, since WorkerModule doesn't directly instantiate any providers, and the sub-modules that require it all import it themselves anyway.

Although, probably worth experimenting--I think this might be a good use case for making this a global module; that way you only have to import it once here & all sub-modules can use its exported services. The way it's set up right now, I think each module that imports it is getting its own instance, which seems unnecessary. That's the pitfall of dependency injection: it makes it a bit too easy to unintentionally instantiate multiple copies of things.

Or maybe I'm wrong; but worth checking.

libs/content-watcher-lib/src/ipfs/ipfs.processor.module.ts
JoeCap08055, line 1: Possibly the subject of a separate PR, but it seems unnecessary to have these queue processors in a "lib" module when they're specific to content-watcher & could just be moved under 'app/content-watcher'.

apps/content-publishing-api/src/controllers/v1/development.controller.v1.ts
aramikm: So this was the only reference to IPFSService used in the content-publishing-api and by removing it we don't need have a dependency on IPFS module. Also this is not an important dev endpoint since we can also get the same info from ipfs dashboard

apps/content-publishing-worker/src/worker.module.ts
aramikm: based on the experiments it looks like we only have 1 instance by default

libs/content-watcher-lib/src/ipfs/ipfs.processor.module.ts
aramikm: yeah, I'll work on that next

JoeCap08055 (approved):

PR #589, opened 2024-10-04
bug: update @polkadot api to work with testnet at 1.13

mattheworris:
# Problem

Gateway containers do not work with Testnet.

# Solution

Update `@polkadot/api@13.2.1` to match Frequency release 1.13.2

## Steps to Verify:

1. Build local containers.
2. Setup SAT to use local containers with Testnet.
3. Login and make a post and confirm that it gets posted to testnet.

## Screenshots (optional):

Screenshot 2024-10-04 at 12 40 05


saraswatpuneet (approved):
Lgtm

PR #593, opened 2024-10-04
content-watcher: move the processors into the app

### LINKED ISSUE #590: Content watcher refactor
# Context
Currently the worker processor are inside watcher-lib. All of these should be moved into the app itself.

aramikm:
# Problem

closes #590 

# changes
- Moved worker processors into the app
- Renamed watcher queues to be different from publishing


docs/graph/webhooks.html
mattheworris, line 1: nit: What causes these html files to get re-generated?

mattheworris (approved):
- Read through changes, pretty straight-forward renaming, moving.
- Non-blocking comment.
🚢 it!

docs/graph/webhooks.html
aramikm: I think I ran the swagger related commands to get the latest changes

PR #597, opened 2024-10-07
add ipfs documentation

### LINKED ISSUE #596: Add Gateway Run Documentation for IPFS
Run -> Deployment -> IPFS

- Lots of links out on how to set it up
- Ingress and Egress
- 3rd party pinning services.

saraswatpuneet:
# Problem

Add a working flow for setting up IPFS and concrete information for devs

Closes #596


mattheworris (approved):
- Read through changes
- Suggested next steps: how to use pinning with gateway services, could be a large effort as I think the current gateway API usage is not standard with Pinata, etc. Infura currently works, but is no longer available as a free service.

🚢 it!

saraswatpuneet:
I think pinning depends on use case and deployment stack of provide but in any case gateway only cares about an endpoint and ipfs gateway url and does not affect the deployment in anyway

Issue #591, opened 2024-10-04
bug: @dsnp/frequency-schemas is out of date and SAT replies are not working.


mattheworris:
### Steps to reproduce

1. Build the latest version of the gateway service containers locally.
2. Setup SAT to use local dev containers and testnet with infura ipfs pinning.
3. Login and post something.
4. Create a reply to a post and see that it fails in `content-publishing-worker`

```bash
2024-10-04 12:46:48 [Nest] 19  - 10/04/2024, 6:46:48 PM   DEBUG [AllExceptionsFilter] {"statusCode":400,"errorReference":"J8qLVoTtDhIw9lZiX2Gs","message":["inReplyTo should be a valid DsnpContentURI!"],"error":"Bad Request"}
2024-10-04 12:46:48 [Nest] 19  - 10/04/2024, 6:46:48 PM   ERROR [AllExceptionsFilter] ErrorReference [J8qLVoTtDhIw9lZiX2Gs] : BadRequestException: Bad Request Exception
```

### Expected behavior

Replies should work correctly.

### Actual behavior

The frontend returns the following error (which follows on from the gateway service error above)

```
ApiException: HTTP-Code: 503
Message: "Error creating broadcast"
    at Context.handleResponse (http://localhost:3000/static/js/bundle.js:26017:11)
    at async createReply (http://localhost:3000/static/js/bundle.js:4255:24)
```

### Debugging notes

`@dsnp/frequency-schemas@1.1.0` had a method `fromFrequencySchema` that was removed sometime after v1.1.0.

[Here](https://github.com/ProjectLibertyLabs/gateway/blob/main/apps/content-publishing-worker/src/batch_announcer/batch.announcer.ts#L46) `content-publishing` references a file, `parquet.ts`, that was removed.

The README.md shows an example usage [here](https://github.com/LibertyDSNP/frequency-schemas/blob/v1.3.2/README.md?plain=1#L56). However, this example does not appear to be correct.

PR #599, opened 2024-10-07
Update and remove dependencies to fix reply issue

wilwade:
# Problem

`@dsnp/frequency-schemas` is being deprecated and changed, so updated to the latest `@dsnp/schemas`

Closes: #591

# Solution

- Removed `@dsnp/frequency-schemas`
- Added `@dsnp/schemas`
- Upgraded various packages


libs/types/src/dtos/account/wallet.login.request.dto.ts
```diff
@@ -7,7 +7,7 @@ import {
   SiwsPayload,
   ValidSignUpPayloads,
   type WalletProxyResponse,
-} from '@projectlibertylabs/siwf';
+} from '@projectlibertylabs/siwfv1';
```
wilwade, line 10: Use the alias version

mattheworris (approved):
🚢 it!

PR #602, opened 2024-10-08
bugfix: fixed content hash

### LINKED ISSUE #601: Fix multi-hash bug
# Context

At appears that the multihash format generated is not correct and `sha256` library already prefixes the hash with multihash value;

# Acceptence criterea
- fix the multihash so that the DsnpContent hash verification works as expected
- Add tests

aramikm:
# Problem

closes #601 

# solution
- had to remove multihash library since the `sha256` library was prefixing the hash algorithm in the first place without us knowing about it

## Verification
- added tests


wilwade (approved):
💯

mattheworris (approved):
- Read through changes
- Tested posts and replies with SAT

🚢 it!

PR #605, opened 2024-10-09
graph-service: add encryption at rest

### LINKED ISSUE #576: Graph Service should ensure that any user Graph keys are encrypted when at rest (aka in the queue)
# Description
As a user, I do not want my private graph encryption keys to be potentially exposed while transiting the `graph-service` workflow.

# Problem
Currently, private graph keys are submitted as part of the REST request and added as part of a job to a queue in BullMQ. SInce BullMQ relies on Redis, this potentially exposes the graph keys to inspection by anyone who gains access to the Redis instance.

# Solution/Acceptance Criteria
- [x] Create a new secret value (seed) to be injected from which an encryption keypair can be generated
- [x] Encrypt the private graph key using the new keypair before it leaves the controller layer
- [x] Decrypt the private graph key only at the point of use

aramikm:
# Problem

closes #576

# Verification
- added test for encryption/decryption

# Steps
- post a graph using graph-api
- Check stored value in redis (private keys should be encrypted)
- Run the graph-worker
- It should be able to decrypt the private key correctly


libs/graph-lib/src/services/encryption.service.ts
```diff
@@ -0,0 +1,123 @@
...
+    }
+  }
+
+  /**
+   * replaces private keys inside GraphKeyPairDto array with an encrypted one
+   * @param graphKeyPairs
+   * @returns the public key hex of encryption key used or returns null if encryption was not required
+   */
+  public async encryptPrivateKeys(graphKeyPairs?: GraphKeyPairDto[]): Promise<EncryptionResult | null> {
+    await this.init();
```
aramikm, line 50: Is there a cleaner way of initializing an async function alternative to what I have here?

aramikm:
@saraswatpuneet Please take a look and let me know if I'm missing the private key encryption anywhere in the graph-service. Thanks

libs/config/src/graph-common.config.ts
```diff
@@ -18,6 +19,10 @@ export default registerAs('graph-common', (): IGraphCommonConfig => {
       value: process.env.GRAPH_ENVIRONMENT_TYPE,
       joi: Joi.string().required().valid('Mainnet', 'TestnetPaseo'),
     },
+    atRestEncryptionKeySeed: {
+      value: process.env.AT_REST_ENCRYPTION_KEY_SEED,
+      joi: Joi.string().required(),
```
JoeCap08055, line 24: q for future: worth adding validation that it's a valid BIP39 seed phrase? @polkadot/util-crypto has a method for this (same for provider account seed phrase, though that one would require a bit of custom validation to allow dev accounts like '//Alice' to pass)

Not important for now; just making a note.

libs/graph-lib/src/services/encryption.service.ts
JoeCap08055: Perhaps better to make `init()` instead be either `onApplicationBootstrap()` or `onModuleInit()`, that way it will get executed before the app enters its main loop, and if you throw from there it will abort app startup.

JoeCap08055: Then any other methods in the class can assume that initialization was complete (or you could layer in other mechanisms if that's not a safe enough assumption)

libs/graph-lib/src/services/encryption.service.ts
```diff
@@ -0,0 +1,123 @@
...
+
+    if (graphKeyPairs !== null && graphKeyPairs.length > 0) {
```
JoeCap08055, line 52: nit: could `graphKeyPairs` be `undefined`? Can be simplified to:
```suggestion
    if (graphKeyPairs?.length) {
```

libs/graph-lib/src/services/encryption.service.ts
```diff
@@ -0,0 +1,123 @@
...
+      const sender = await this.suite.createSenderContext({
+        recipientPublicKey: this.encryptionKey.publicKey,
+      });
+
+      const decryptedKeyPairs: GraphKeyPairDto[] = [];
+      for (let i = 0; i < graphKeyPairs.length; i += 1) {
+        const keyPair = lodash.cloneDeep(graphKeyPairs[i]);
```
JoeCap08055, line 59: ~Taking a dependency on lodash is a hot-button topic... do we need to? It adds a lot of bloat to the dependency tree.~
nvm--looks like NestJS already adds a dependency on lodash ☹️

JoeCap08055: Or better, use native ES6 `structuredClone`:
```suggestion
        const keyPair = structuredClone(graphKeyPairs[i]);
```

libs/graph-lib/src/services/encryption.service.ts
```diff
@@ -0,0 +1,123 @@
...
+   */
+  public async encryptPrivateKeys(graphKeyPairs?: GraphKeyPairDto[]): Promise<EncryptionResult | null> {
+    await this.init();
+
+    if (graphKeyPairs !== null && graphKeyPairs.length > 0) {
+      const sender = await this.suite.createSenderContext({
+        recipientPublicKey: this.encryptionKey.publicKey,
+      });
+
+      const decryptedKeyPairs: GraphKeyPairDto[] = [];
```
JoeCap08055, line 57:
```suggestion
      const encryptedKeyPairs: GraphKeyPairDto[] = [];
```

libs/graph-lib/src/services/encryption.service.ts
```diff
@@ -0,0 +1,123 @@
...
+   * @param graphKeyPairs
+   */
+  public async decryptPrivateKeys(
+    encryptionPublicKeyHex: string | undefined,
+    senderContextHex: string | undefined,
+    graphKeyPairs?: GraphKeyPairDto[] | undefined,
+  ): Promise<GraphKeyPairDto[] | null> {
+    await this.init();
+
+    if (graphKeyPairs !== undefined && graphKeyPairs !== null && graphKeyPairs.length > 0) {
```
JoeCap08055, line 88:
```suggestion
    if (graphKeyPairs?.length) {
```

libs/graph-lib/src/services/encryption.service.ts
```diff
@@ -0,0 +1,123 @@
...
+   * replaces private keys inside GraphKeyPairDto array with an encrypted one
+   * @param graphKeyPairs
+   * @returns the public key hex of encryption key used or returns null if encryption was not required
+   */
+  public async encryptPrivateKeys(graphKeyPairs?: GraphKeyPairDto[]): Promise<EncryptionResult | null> {
+    await this.init();
+
+    if (graphKeyPairs !== null && graphKeyPairs.length > 0) {
+      const sender = await this.suite.createSenderContext({
+        recipientPublicKey: this.encryptionKey.publicKey,
```
JoeCap08055, line 54: Do we need `senderKey` here as well?

libs/config/src/graph-common.config.ts
aramikm: This seed can be anything. I don't think it has to be a BIP39 since we are using it in key derivation for X25519 keys which is used for encryption

libs/graph-lib/src/services/encryption.service.ts
aramikm: I was looking for an alternative, thanks Was kinda iffy adding lodash just because of this.

libs/graph-lib/src/services/encryption.service.ts
aramikm: In encryption we are generating a new key by calling `createSenderContext` this key is used for diffie-hillman with the `encryptionKey` which generates a secret that is used for AES256GCM algorithm to encrypt

apps/graph-api/src/api.service.ts
```diff
@@ -79,7 +79,7 @@ export class ApiService implements BeforeApplicationShutdown {
     const job = await this.graphChangeRequestQueue.add(`Request Job - ${data.referenceId}`, data, {
       jobId: data.referenceId,
     });
-    this.logger.debug(job);
+    this.logger.debug(JSON.stringify(job));
```
JoeCap08055, line 82: Sometimes `this.logger.debug('Some message', obj)` works without having to invoke JSON.stringify. Worth a test.

developer-docs/graph/ENVIRONMENT.md
```diff
@@ -4,7 +4,7 @@
 This application recognizes the following environment variables:
 
 | Name                               | Description                                                                                                                                                                       |                                               Range/Type                                               | Required? |     Default      |
-| ---------------------------------- |-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------| :----------------------------------------------------------------------------------------------------: |:---------:|:----------------:|
+| ---------------------------------- | --------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- | :----------------------------------------------------------------------------------------------------: | :-------: | :--------------: |
```
JoeCap08055, line 7: Cool--never knew about the `:` aligment trick

libs/graph-lib/src/services/encryption.service.ts
```diff
@@ -17,53 +16,49 @@ export class EncryptionResult {
...
-  private readonly suite: CipherSuite;
+  private suite: CipherSuite;
 
   private encryptionKey: CryptoKeyPair;
 
   constructor(@Inject(graphCommonConfig.KEY) private readonly config: IGraphCommonConfig) {
     this.logger = new Logger(this.constructor.name);
+  }
+
+  async onModuleInit() {
```
JoeCap08055, line 30: `this.suite` can still be `private readonly` and initialized in constructor, since it doesn't need any async initialization.

JoeCap08055 (approved):
One or two nits, but nothing important/blocking.
Great work!

mattheworris (approved):
- Read through changes
🚢 it

apps/graph-api/src/api.service.ts
aramikm: I think this was working before we added `transform: true,` this changes the raw javascript object into the object and somehow it prints object when logged directly

PR #608, opened 2024-10-09
blockchain: event errors

### LINKED ISSUE #500: Correct error handling for blockchain-service when an expected event is not found.
### Context
 Currently it doesn't looks like we are handling the scenario that the expected event might not exist on chain.

example code
```typescript
  public handlePublishKeyTxResult(event: Event): PublicKeyValues {
    const publicKeyValues: Partial<PublicKeyValues> = {};

    // Grab the event data
    if (event && this.api.events.msa.PublicKeyAdded.is(event)) {
      publicKeyValues.msaId = event.data.msaId.toString();
      publicKeyValues.newPublicKey = event.data.key.toString();
      publicKeyValues.debugMsg = `Public Key: ${publicKeyValues.newPublicKey} Added for msaId: ${publicKeyValues.msaId}`;
    }

    return publicKeyValues as PublicKeyValues;
  }
```

In the above case and other ones existing in that file if the event does not exist we will have an unhandled edge case.

aramikm:
# Problem

closes #500

# Description
- Was going to throw error but it didn't help since it would cause the block scanning process to be stuck on that certain block and would not be able to recover. The better option seemed like to only log these cases.

### DIFF
--- a/libs/blockchain/src/blockchain-rpc-query.service.ts
+++ b/libs/blockchain/src/blockchain-rpc-query.service.ts
@@ -467,6 +467,7 @@ export class BlockchainRpcQueryService implements OnApplicationBootstrap, OnAppl
       };
     }
 
+    this.logger.error(`Expected HandleClaimed event but found ${event}`);
     return {} as HandleTxnValues;
   }
 
@@ -488,6 +489,7 @@ export class BlockchainRpcQueryService implements OnApplicationBootstrap, OnAppl
       };
     }
 
+    this.logger.error(`Expected PublicKeyAdded event but found ${event}`);
     return {} as PublicKeyValues;
   }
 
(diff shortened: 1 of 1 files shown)


libs/blockchain/src/blockchain-rpc-query.service.ts
```diff
@@ -467,6 +467,7 @@ export class BlockchainRpcQueryService implements OnApplicationBootstrap, OnAppl
       };
     }
 
+    this.logger.error(`Expected HandleClaimed event but found ${event}`);
```
aramikm, line 470: Was going to throw error but it didn't help since it would cause the block scanning process to be stuck on that certain block and would not be able to recover. The better option seemed like to only log these cases. Open to other suggestions.

mattheworris (approved):
🚢 it!

PR #628, opened 2024-10-18
graph-service: added connection type to request

### LINKED ISSUE #627: [Graph-service]: some improvements and fixes
# Description

After our major refactoring there might be some issues that we will address in this ticket.

aramikm:
# Problem

closes #627 

# Changes
- fixed the edge case that when we wanted to get public graph but in the request we passed incorrect private keys, we are going to ignore the private keys and continue with fetching the public graph
- Added missing connectionType to the `POST /v1/graphs/getGraphs` request
- Added a errorMessage in the `POST /v1/graphs/getGraphs` response so that we can expose the error to the user instead of returning empty graph

# Impact
- We will need to ensure these changes will get reflected in `social-app-template `

## Screenshots (optional):

Screenshot 2024-10-18 at 4 54 48 PM
Screenshot 2024-10-18 at 4 55 03 PM

Error response example:
Screenshot 2024-10-18 at 4 56 20 PM


saraswatpuneet (approved):
lgtm

libs/graph-lib/src/services/graph-state-manager.ts
```diff
@@ -115,6 +115,17 @@ export class GraphStateManager implements OnApplicationBootstrap {
...
 
+  public getConnectionTypeForSchema(schemaId: number): ConnectionType {
+    let connectionType = ConnectionType.Friendship;
+    if (
+      this.schemaIds[ConnectionType.Follow][PrivacyType.Private] === schemaId ||
+      this.schemaIds[ConnectionType.Follow][PrivacyType.Public] === schemaId
+    ) {
+      connectionType = ConnectionType.Follow;
+    }
+    return connectionType;
```
JoeCap08055, line 126: Just some code golf if you want a more generic way of doing this; not really necessary since there's only 2 entries at each level here, but if you're interested:
```suggestion
return Object.entries(this.schemaId).filter(([connectionType, v]) => Object.values(v).includes(schemaId)).map(([connectionType, v]) => connectionType)[0] || ConnectionType.Friendship;
```

Frankly, yours is more readable, but I still thought this was cool.

JoeCap08055 (approved):
Looks good! Just some code golf musing from me; feel free to ignore. Approved.

libs/graph-lib/src/services/graph-state-manager.ts
aramikm: Thanks for the suggestion. I think it would make it harder to read and understand.

PR #638, opened 2024-10-21
chore(docs): update graph service

enddynayn:
Update Graph Service documentation so that
it reflects the latest changes.

### DIFF
--- a/developer-docs/graph/README.md
+++ b/developer-docs/graph/README.md
@@ -1,285 +1,163 @@
 # Graph Service
 
-<!-- TABLE OF CONTENTS -->
+The Graph Service is a crucial component of the Gateway suite, enabling easy interaction with DSNP private and public graphs on the Frequency blockchain. This document provides an overview of the service, its architecture, and guides for setup and usage.
 
-# 📗 Table of Contents
+## 📗 Table of Contents
 
 - [📖 About the Project](#about-project)
-- [🔍 Arch Map](#-arch-maps)
-- [🛠 Built With](#-built-with)
-  - [Tech Stack](#tech-stack)
-  - [Key Features](#key-features)
-- [🚀 Live OpenAPI Docs](#-live-docs)
-- [💻 Getting Started](#-getting-started)
-  - [Prerequisites](#prerequisites)
-  - [Setup](#setup)
-  - [Install](#install)
-  - [Usage](#usage)
-  - [Run tests](#run-tests)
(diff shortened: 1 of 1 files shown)


claireolmstead (approved):
Read through it and it looks good to me!!! 🚢

developer-docs/graph/README.md
```diff
@@ -1,285 +1,155 @@
...
+3. Set up environment variables:
+   ```bash
+   cp env-files/graph.template.env .env
+   ```
+   Edit the `.env` file with appropriate values for your environment.
 
-Note: using [docker compose file](docker-compose.yaml) to start the services. This will start the services in development mode.
+4. Start all services using Docker Compose:
+   ```bash
+   docker compose up -d redis frequency graph-service-api graph-service-worker
```
aramikm, line 67: I don't think works. It looks like we've 2 instances of redis
- redis
- bull

I'm unaware why we've split them in the first place but that might be the issue

aramikm: you might also need to run the bull instance on the requested port

aramikm: Actually this seem to work the problem was that when you don't have a registered provider the worker just doesn't error out and just hangs there and that's something that I think we have a ticket for

developer-docs/graph/README.md
```diff
@@ -1,285 +1,155 @@
...
+   ```bash
+   docker compose up -d redis frequency
+   ```
 
-```sh
-make setup
-```
+2. Set up the graph with test data:
+   ```bash
+   make setup-account
```
aramikm, line 81: I think you would need to run this in both cases (locally and docker) to have a registered provider with capacity.

developer-docs/graph/README.md
```diff
@@ -1,285 +1,155 @@
...
+4. Start the Graph Service Worker:
+   ```bash
+   make start-graph-worker
+   ```
 
-#### 4. Check the job in [BullUI](http://0.0.0.0:3000/queues/), to monitor job progress based on defined tests.
+### Verification
 
-## 📋 Testing
+- Access Swagger UI: [http://localhost:3000/docs/swagger](http://localhost:3000/docs/swagger)
```
aramikm, line 96: the ports would be different when we run it inside docker

enddynayn: Good catch. I updated the documentation.

PR #642, opened 2024-10-22
Chore/handle chain rpc node disconnect

### LINKED ISSUE #473: Handle Blockchain RPC Failures
If the RPC node goes down, how does Gateway handle it?

## Acceptance Criteria

- [x] All Services react to a down RPC in an expected way
- [x] Good logging when RPC is disconnected

### Implementation Options (might be different for different services)
- Have the ability for more than one node and try to failover to a different node
- Wait a timeout and exit the application when it cannot reconnect to the RPC
- Go into a degraded "pending" state where it can still collect data, but doesn't submit them to the chain (content publisher for example)

### Runtime Upgrade Test

While running Gateway against testnet, testnet did a runtime upgrade which triggered a restart of the services. Assuming k8 restart, this correctly solves the runtime upgrade issue, but may want more.

JoeCap08055:
# Description
This PR accomplishes the following:
- Refactor all connection-related logic out of `BlockchainRpcQueryService` into a new base class, `PolkadotApiService`
    - `PolkadotApiService` emits events 'chain.connected' and 'chain.disconnected'
-  Handle chain connect/disconnect with a timeout that terminates the app if the chain connection is gone for some amount of time
- Add a new config, `FREQUENCY_TIMEOUT_SECS` (default: 10 seconds)
- In blockchain scanners, suspend scan if chain is disconnected; resume when chain is connected
- Rework mocking of Polkadot API in unit tests to accommodate new class structure

Closes #473


apps/account-api/src/services/accounts.service.ts
```diff
@@ -70,7 +71,7 @@ export class AccountsService {
 
   // eslint-disable-next-line class-methods-use-this
   async signInWithFrequency(request: WalletLoginRequestDto): Promise<WalletLoginResponseDto> {
-    const api = await this.blockchainService.getApi();
+    const api = (await this.blockchainService.getApi()) as ApiPromise;
```
aramikm, line 74: question: why do we need this type casting?

apps/account-worker/src/transaction_publisher/publisher.service.ts
```diff
@@ -188,7 +188,7 @@ export class TransactionPublisherService extends BaseConsumer implements OnAppli
     try {
       const prefixedSignature: SignerResult = { id: 1, signature };
       const signer: Signer = getSignerForRawSignature(prefixedSignature);
-      const { nonce } = await this.blockchainService.api.query.system.account(accountId);
+      const nonce = await this.blockchainService.getNonce(accountId);
```
aramikm, line 191: great find!

libs/blockchain/src/polkadot-api.service.ts
```diff
@@ -0,0 +1,136 @@
...
+        this.provider = new HttpProvider(this.baseConfig.frequencyApiWsUrl.toString());
+      } else {
+        throw new Error('Unrecognized chain URL type', { cause: providerUrl.toJSON() });
+      }
+
+      this.api = new ApiPromise({ ...options, provider: this.provider, noInitWarn: false });
+      // From https://github.com/polkadot-js/api/blob/700812aa6075c85d8306451ce062d8c06b161e2b/packages/api/src/promise/Api.ts#L157C1-L159C1
+      // Swallow any rejections on isReadyOrError
+      // (in Node 15.x this creates issues, when not being looked at)
+      this.api.isReadyOrError.catch(() => {});
```
aramikm, line 66: nit: I'm curious on how this was found

aramikm (approved):
Looks great! You might need to regenerate the swagger or some types.

libs/blockchain/src/polkadot-api.service.ts
JoeCap08055: Painful experience & Google 😄

JoeCap08055: Also digging into polkadot-js code

apps/account-api/src/services/accounts.service.ts
JoeCap08055: Because `getApi()` can return one of 2 things, depending on whether you pass in an optional block hash argument: either the base `ApiPromise` -or- the result of `api.at(blockHash)`, which is not an ApiPromise, but is instead `ApiDecoration<'promise'>`. However, the latter is only valid for RPC calls and state queries, not signing stuff or creating extrinsics, I believe.

Until the SIWFv2 stuff went in, this wasn't really an issue... I'm not a fan of type-casting in most cases, but here it seems pretty inocuous.

libs/blockchain/src/blockchain.config.ts
```diff
@@ -76,6 +77,10 @@ const doRegister = (mode: ChainMode = ChainMode.PROVIDER_SEED_REQUIRED) =>
     }
 
     const configs: JoiUtil.JoiConfig<IBlockchainConfig> = {
+      frequencyTimeoutSecs: {
+        value: process.env.FREQUENCY_TIMEOUT_SECS,
+        joi: Joi.number().positive().default(60),
```
mattheworris, line 82: nit: Here the default is 60, the environment template files set a value of 10. Do we want these values to be in sync?

Do we want to update the docker-compose.yaml files to get/set FREQUENCY_TIMEOUT_SECS?

JoeCap08055: Good catch. I had set it to 10 in the environment to make testing easier, but that's probably too short. I'll bump it back to 60 in one of my other pending PRs.

mattheworris (commented):
- Read through changes 
- Ran some tests with social-app-template
- Nice work!

PR #644, opened 2024-10-22
chore(docs): update content-watcher docs

enddynayn:
Update Content-Watcher-Service Documentation structure
and content.

### DIFF
--- a/developer-docs/content-watcher/README.md
+++ b/developer-docs/content-watcher/README.md
@@ -1,356 +1,189 @@
-# Content Watcher
+# Content Watcher Service
 
-<!-- TABLE OF CONTENTS -->
+The Content Watcher Service is a crucial component of the Gateway suite, providing a familiar callback API to retrieve content and publishing announcements from the Frequency blockchain. This document provides an overview of the service, its architecture, and guides for setup and usage.
 
-# 📗 Table of Contents
+## 📗 Table of Contents
 
 - [📖 About the Project](#about-project)
-- [🔍 Arch Map](#-arch-maps)
-- [🛠 Built With](#-built-with)
-  - [Tech Stack](#tech-stack)
-  - [Key Features](#key-features)
-- [🚀 Live OpenAPI Docs](#-live-docs)
-- [💻 Getting Started](#-getting-started)
-  - [Prerequisites](#prerequisites)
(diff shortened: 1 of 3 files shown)


saraswatpuneet (approved):
Much cleaner 🧹

developer-docs/content-watcher/README.md
```diff
@@ -1,356 +1,186 @@
...
-- **Send content to registered webhooks**
-- **Simple Schema and MSA Id based filtering**
-
-#### Scanner API
-
-- **Start, Stop, and restart scanning**
+## 📖 About the Project <a name="about-project"></a>
 
-<p align="right">(<a href="#-table-of-contents">back to top</a>)</p>
+The Content Watcher Service is part of the Gateway suite that provides a Web2-friendly interface for monitoring and retrieving blockchain content. It implements a webhook system that pushes updates to registered endpoints, making it easy to stay synchronized with chain events.
```
claireolmstead, line 19: Do we want to link to gateway?

developer-docs/content-watcher/README.md
```diff
@@ -1,356 +1,186 @@
...
 
-<!-- LIVE Docs -->
+## 🔍 Architecture Overview <a name="architecture-overview"></a>
 
-## 🚀 Live Docs
+The Content Watcher Service provides a webhook-based system for receiving notifications about new content. Even with a crawl request, results are pushed to the webhook rather than retrieved directly.
```
claireolmstead, line 23: The second sentence is a little confusing to me. Thinking either we drop it or explain why we cannot retrieve it directly.

claireolmstead (commented):
I think the big thing that this is missing is an explanation as to why we have to watch content, which should be near the top. We have to watch becuase things are happening async due to the nature of the chain and how we're queueing posts and puts on chain. I would like to see a little more of a technical description around that.

enddynayn:
> I think the big thing that this is missing is an explanation as to why we have to watch content, which should be near the top. We have to watch becuase things are happening async due to the nature of the chain and how we're queueing posts and puts on chain. I would like to see a little more of a technical description around that.

Good point. Thanks! I update the documentation to reflect your suggestion.

claireolmstead (approved):
Great changes!

PR #648, opened 2024-10-23
fix: make sure correct tag is used when release workflow is triggered manually

JoeCap08055:
This PR fixes an issue where the release workflow processes a blank tag when triggered manually instead of from a release event.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -35,7 +35,7 @@ jobs:
           test_run=false
           ENTIRE_TAG_RE="^(([[:alpha:]\-]+)-)?(.*)$"
           VERSION_TAG_RE="^v[\.]?([[:digit:]]+\.[[:digit:]]+\.[[:digit:]]+(-[[:alpha:][:digit:]\.]+)*)$"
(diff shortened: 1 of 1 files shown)


saraswatpuneet (approved):

wilwade (approved):
Looks correct to me.

.github/workflows/release.yml
```diff
@@ -35,7 +35,7 @@ jobs:
           test_run=false
           ENTIRE_TAG_RE="^(([[:alpha:]\-]+)-)?(.*)$"
           VERSION_TAG_RE="^v[\.]?([[:digit:]]+\.[[:digit:]]+\.[[:digit:]]+(-[[:alpha:][:digit:]\.]+)*)$"
-          TAG="${{ github.event.release.tag_name }}"
+          TAG="${FULL_TAG}"
```
mattheworris, line 38: nit: not sure of all the syntactic differences, but this is what chatGPT suggested and I was able to test with successfully yesterday
```suggestion
          TAG="${{env.FULL_TAG}}"
```

mattheworris (approved):

PR #663, opened 2024-10-25
fix: use stock Frequency node image

JoeCap08055:
Switches from `dsnp/instant-seal-node-with-deployed-schemas` to `frequencychain/stanadlone-node` for Frequency node Docker image, now that the standard image includes the Mainnet schemas in the genesis block. (Should result in a speedier node startup, as we do not need to wait for the schemas to be deployed by an embedded script)

Also added a health check command to the container so that any services that depend on it won't come up until the RPC node is ready to serve requests.


wilwade (approved):
🎉

mattheworris (approved):
🚢 it!

saraswatpuneet (approved):
Good stuff

docker-compose.yaml
```diff
@@ -84,6 +74,11 @@ services:
     # such as:
     # --state-pruning=archive
     command: --offchain-worker=always --enable-offchain-indexing=true
+    options: >-
+      --health-cmd "bash -c '(echo >\"/dev/tcp/127.0.0.1/9944\") &>/dev/null'"
+      --health-interval 10s
+      --health-timeout 5s
+      --health-retries 5
```
mattheworris, line 81:
```suggestion
    healthcheck:
      test: ["CMD", "bash", "-c", '(echo >\"/dev/tcp/127.0.0.1/9944\") &>/dev/null']
      interval: 10s
      timeout: 5s
      retries: 5
```

PR #664, opened 2024-10-25
feat: add webhook support for retireMsa, revokeDelegation

### LINKED ISSUE #617: Account Service should handle webhook notifications for `RETIRE_MSA` and `REVOKE_DELEGATION`
As a developer, I expect that when a `RETIRE_MSA` or `REVOKE_DELEGATION` extrinsic is submitted, that the notifier will look for the transaction success events and send a notification to the webhook, if registered.

### Solution

Make sure this [switch case](https://github.com/ProjectLibertyLabs/gateway/blob/main/apps/account-worker/src/transaction_notifier/notifier.service.ts#L147) also handles `TransactionType.RETIRE_MSA` and `TransactionType.REVOKE_DELEGATION`.

### Alternate Solution

If it is decided that proxy transactions are not to be monitored and no webhook notifications are to be sent, do not add these transactions to the watch cache [here](https://github.com/ProjectLibertyLabs/gateway/blob/main/apps/account-worker/src/transaction_publisher/publisher.service.ts#L139)

JoeCap08055:
# Problem

`retireMsa` and `revokeDelegation` extrinsic submissions were not properly linked to the webhook processing, hence no webhook notifications were sent. This PR adds the hooks in to the webhook processing so that notifications may be sent when these operations complete.

Closes #617


mattheworris (approved):
- Read through changes
🚢 it!

PR #665, opened 2024-10-25
fix: healthcheck specification in docker-compose file

JoeCap08055:
# Problem

Healthcheck was specified in docker-compose.yaml using incorrect syntax.
Odd that CI didn't catch the error...


mattheworris (approved):
🚢 it!

PR #673, opened 2024-11-04
Check capacity enough for tx

### LINKED ISSUE #670: [bug] account can have insufficient capacity even if within limits
# Description
When submitting transactions to the chain, typically we check first to see if available capacity is within configured tolerances; if not, we return the job to the queue & pause the queue until more capacity is available (either at the next epoch, or after a new stake)

However, having available capacity within configured tolerances does not guarantee that there is enough for any specific transaction, especially if the capacity tolerance limits are not properly tuned for the amount of capacity that a provider has. In such as case, the transaction would reject with an RPC error, and, according to current logic, not be returned to the queue, but instead fail the job.

It might be prudent to also check the stable weight of the candidate transaction against available capacity before submitting to the chain as well.

saraswatpuneet:
# Problem
 Closes #670 

# Solution

Before submitting capacity transaction via `payWithCapacity` or `payWithCapacityBatchAll` check if tx fee in capacity is enough as compared for remaining capacity from provider

# Tests
- [x] Test full e2e and checks are working


libs/blockchain/src/blockchain.service.ts
```diff
@@ -169,6 +174,12 @@ export class BlockchainService extends BlockchainRpcQueryService implements OnAp
   public async payWithCapacityBatchAll(
     tx: Vec<Call> | (Call | IMethod | string | Uint8Array)[],
   ): Promise<[SubmittableExtrinsic<'promise'>, HexString]> {
+    tx.forEach(async (call) => {
+      const outOfCapacity = await this.checkTxCapacityLimit(this.config.providerId, call.toString());
+      if (outOfCapacity) {
+        throw new DelayedError();
+      }
+    });
```
saraswatpuneet, line 182: do we want to add up each call capacity and compare against remaining?

libs/blockchain/src/blockchain.service.ts
```diff
@@ -154,6 +155,10 @@ export class BlockchainService extends BlockchainRpcQueryService implements OnAp
   public async payWithCapacity(
     tx: Call | IMethod | string | Uint8Array,
   ): Promise<[SubmittableExtrinsic<'promise'>, HexString]> {
+    const outOfCapacity = await this.checkTxCapacityLimit(this.config.providerId, tx.toString());
+    if (outOfCapacity) {
+      throw new DelayedError();
+    }
```
saraswatpuneet, line 161: if there isnt enough capacity for tx then throw DelayedError that should be handled at caller end

apps/graph-worker/src/graph_publisher/graph.publisher.processor.service.ts
```diff
@@ -118,6 +118,9 @@ export class GraphUpdatePublisherService extends BaseConsumer implements OnAppli
 
       this.logger.debug('Cached extrinsic to monitor: ', payWithCapacityTxHash);
     } catch (error: unknown) {
+      if (error instanceof DelayedError) {
+        job.moveToDelayed(Date.now(), job.token); // fake delay, we just want to avoid processing the current job if we're out of capacity
+      }
```
saraswatpuneet, line 123: if DelayedError is caught then delay and throw

apps/account-worker/src/transaction_publisher/publisher.service.ts
```diff
@@ -67,7 +67,6 @@ export class TransactionPublisherService extends BaseConsumer implements OnAppli
     try {
       // Check capacity first; if out of capacity, send job back to queue
       if (!(await this.capacityCheckerService.checkForSufficientCapacity())) {
-        job.moveToDelayed(Date.now(), job.token); // fake delay, we just want to avoid processing the current job if we're out of capacity
```
saraswatpuneet, line 70: moved it to outer scope

apps/account-worker/src/transaction_publisher/publisher.service.ts
```diff
@@ -138,10 +137,11 @@ export class TransactionPublisherService extends BaseConsumer implements OnAppli
       obj[txHash] = JSON.stringify(status);
       this.cacheManager.hset(TXN_WATCH_LIST_KEY, obj);
     } catch (error: any) {
-      if (!(error instanceof DelayedError)) {
+      if (error instanceof DelayedError) {
+        job.moveToDelayed(Date.now(), job.token);
+      } else {
```
saraswatpuneet, line 142: if DelayedError is caught then delay and throw

libs/blockchain/src/blockchain-rpc-query.service.ts
```diff
@@ -237,6 +245,23 @@ export class BlockchainRpcQueryService extends PolkadotApiService {
...
+    if (capacityCost.inclusionFee.isSome) {
+      const inclusionFee = capacityCost.inclusionFee.unwrap();
+      totalAdjustedWeightFee += inclusionFee.adjustedWeightFee.toBigInt();
+    }
+
+    // Check if the total cost exceeds the remaining capacity
+    const outOfCapacity = remainingCapacity < totalAdjustedWeightFee;
+    return outOfCapacity;
+  }
+
```
saraswatpuneet, line 264: this calls `rpc` to get `FeeDetails` which has adjusted fee that should be less than remaining capacity. this is not a breaking change if rpc fails for some bad transaction and returns null this logic will default to false

libs/blockchain/src/blockchain-rpc-query.service.ts
```diff
@@ -237,6 +245,23 @@ export class BlockchainRpcQueryService extends PolkadotApiService {
...
+    const capacityInfo = await this.capacityInfo(providerId);
+    const { remainingCapacity } = capacityInfo;
+
+    // Calculate the total capacity cost for all encoded extensions
+    let totalAdjustedWeightFee = BigInt(0);
+    const capacityCost = await this.getCapacityCostForExt(encodedExt);
+    if (capacityCost.inclusionFee.isSome) {
+      const inclusionFee = capacityCost.inclusionFee.unwrap();
+      totalAdjustedWeightFee += inclusionFee.adjustedWeightFee.toBigInt();
+    }
```
JoeCap08055, line 258: Should maybe put this in a try/catch so a call for an invalid extrinsic can return false instead of throw an error

saraswatpuneet: Good point, thought the rpc handled wrong input and return free detail will none but will put a try catch

libs/blockchain/src/blockchain-rpc-query.service.ts
```diff
@@ -252,6 +277,10 @@ export class BlockchainRpcQueryService extends PolkadotApiService {
     return (await api.query.capacity.epochLength()).toNumber();
   }
 
+  public async getCapacityCostForExt(enocdedExt: HexString): Promise<FeeDetails> {
+    return this.api.rpc.frequencyTxPayment.computeCapacityFeeDetails(enocdedExt, null);
+  }
+
```
JoeCap08055, line 283: I've been trying to isolate clients of this class from having to decode SCALE; basically redefining all interfaces in terms of regular JS types and exporting those interfaces at the top of the file. So you would export an `IFeeDetails` interface. Each method is responsible for converting its results and returning them--either returning `value | null`, or `value` (with some default for an empty response), or throwing an error, as appropriate.
 
 Up to you, but that's the pattern this class was developed with. Also makes it much easier to mock in unit tests. See the `capacityInfo()` method and corresponding `ICapacityInfo` interface for an example.

saraswatpuneet: I like that design choice, will follow through the example :+1:

JoeCap08055: Sounds good--while you're at it, I suppose it's worth putting a comment header at the top of this file explaining that design choice for future maintainers.

JoeCap08055: 👍🏻 
Mind putting in a file header comment like I mentioned above? Probably should go in all 3 blockchain class files: `PolkadotApiService`, `BlockchainRpcQueryService`, and `BlockchainService`. Maybe:

```typescript
/*
 * NOTE: This class is designed to isolate consumers from having to deal with the details of interacting directly
 *       with the Frequency blockchain. To that end, return values of functions should not expose the SCALE-
 *       encoded objects that are directly returned from Frequency RPC calls; rather, all payloads should be
 *       unwrapped and re-formed using native Javascript types.
 *
 *       RPC methods that have the potential to return values wrapped as `Option<...>` or any value supporting
 *       the `.isEmpty`, `.isSome`, or `.isNone` getters should implement one of the following behaviors:
 *          - have a specified return type of `<type> | null` and return null for an empty value
 *          - return some sane default for an empty value
 *          - throw an error if an empty value is encountered
 */
 ```

File: apps/content-publishing-worker/src/publisher/publishing.service.ts [hunk 9]
```diff
@@ -56,7 +56,6 @@ export class PublishingService extends BaseConsumer implements OnApplicationBoot
     try {
       // Check capacity first; if out of capacity, send job back to queue
       if (!(await this.capacityCheckerService.checkForSufficientCapacity())) {
-        job.moveToDelayed(Date.now(), job.token); // fake delay, we just want to avoid processing the current job if we're out of capacity
```
mattheworris, line 59: Does this need to be moved to the outer scope like the change in account-worker above?

saraswatpuneet: good catch, i missed this

apps/graph-worker/src/graph_publisher/graph.publisher.processor.service.ts
```diff
@@ -118,6 +119,9 @@ export class GraphUpdatePublisherService extends BaseConsumer implements OnAppli
 
       this.logger.debug('Cached extrinsic to monitor: ', payWithCapacityTxHash);
     } catch (error: unknown) {
+      if (error instanceof DelayedError) {
```
mattheworris, line 122: `process()` is not calling `capacityCheckerService` above (like the other services), should it be?

saraswatpuneet: no the payWithCapacity and payWithCapacityBatchAll functions internally do that check do consumers dont have to worry about it

saraswatpuneet: this is mostly checking capacity only, but can be used else though the current implementation wants to make sure a tx is only sent if there is enough capacity

JoeCap08055 (commented):
I know we don't have a good enough e2e test framework to easily add this to CI, but were you able to engineer a manual test scenario that creates the capacity edge case & test locally that this solves it correctly?

mattheworris (approved):
- Read through changes
- Nice work!
🚢 it!

saraswatpuneet:
@JoeCap08055 read your comment late, I did little bit of manual testing at the edge, I don't think that covers,

I plan to write e2e tests but need some familiarity with them first so a PR will come next week

PR #677, opened 2024-11-05
feat: add handle validation using blockchain service in HandlesControllerV1 and AccountsControllerV2

### LINKED ISSUE #676: Gateway should validate handle before adding the transaction to the queue and returning 201
### Steps to reproduce

1. Use SIWF v2 to create a new user.
2. Request a handle that is longer than 20 characters. **NOTE:** This is time sensitive, the SIWF UI will be updated not to allow invalid lengths.
3. Observe that the claim handle transaction fails.

### Expected behavior

The `account-service-api` should validate the handle using the `validateHandle` RPC before queueing any transactions.

If the handle is invalid return a 400 BAD REQUEST.

mattheworris:
# Problem
When working with handles, new handles should be validated before sending transactions to the blockchain.

Closes #676 

# Solution

Implemented `isValidHandle()` which uses `validateHandle()` RPC to validate base handles before sending them to the blockchain.

## Steps to Verify:

1. Setup Account Service.
2. Use Swagger to test `/v1/handles/change/{newHandle}` a base handle that is longer than 20 characters and verify that the response is 400.
3. Use example.mjs to use SIWF v2 and choose a handle that is too long and observe the 400 response.

Screenshot 2024-11-05 at 10 19 05

### DIFF
--- a/.github/workflows/e2e-tests.yml
+++ b/.github/workflows/e2e-tests.yml
@@ -70,13 +70,12 @@ jobs:
           echo "CHANGES=${{ needs.service-matrix.outputs.changes }}"
           echo "SERVICES=${{ needs.service-matrix.outputs.services }}"
           echo "RUN-ALL=${{ needs.service-matrix.outputs.run-all }}"
-          echo "ACCOUNT_CHANGE_DETECTED=${{ contains(fromJson(needs.service-matrix.outputs.changes), 'account') }}"
+          echo "ACCOUNT_CHANGE_DETECTED=${{ contains(fromJson(needs.service-matrix.outputs.changes), matrix.service) }}"
 
       - name: Run or Skip
         id: should
         run: |
-          echo "MATRIX_SERVICE=${{ matrix.service }}"
-          echo "RUN=${{contains(fromJson(needs.service-matrix.outputs.changes), matrix.service) }}" >> "$GITHUB_OUTPUT"
+          echo "RUN=${{ needs.service-matrix.outputs.run-all || contains(fromJson(needs.service-matrix.outputs.changes), matrix.service) }}" >> "$GITHUB_OUTPUT"
 
       - name: Checkout
         if: ${{ steps.should.outputs.RUN == 'true' }}
--- a/.github/workflows/e2e-tests.yml
+++ b/.github/workflows/e2e-tests.yml
@@ -53,6 +53,8 @@ jobs:
               - 'apps/content-watcher/**'
               - 'libs/content-watcher-lib/**'
               - 'libs/storage/**'
+      - name: Debug determine-matrix changes output
+        run: echo 'CHANGES=${{ steps.determine-matrix.outputs.changes }}'
 
   build:
     name: '[${{ matrix.service }}] E2E Tests'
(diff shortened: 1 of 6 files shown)


libs/account-lib/src/utils/blockchain-scanner.service.ts
```diff
@@ -89,7 +89,7 @@ export abstract class BlockchainScannerService {
         this.scanInProgress = false;
         return;
       }
-      this.logger.verbose(`Starting scan from block #${currentBlockNumber}`);
+      // this.logger.verbose(`Starting scan from block #${currentBlockNumber}`);
```
mattheworris, line 92: These log messages are filling the log and don't generally seem to be useful.
Is it worthwhile to try and filter it somehow and log every 5-10 minutes or something like that?

apps/account-api/test/handles.controller.e2e-spec.ts
```diff
@@ -169,5 +169,13 @@ describe('Handles Controller', () => {
         .expect(HttpStatus.NOT_FOUND)
         .expect((res) => expect(res.text).toContain('Invalid msaId'));
     });
+
+    it('(GET) /handles/change/:newHandle with invalid handle', async () => {
```
mattheworris, line 173: This endpoint is used to build the payload to test the other endpoints.

apps/account-api/src/controllers/v1/handles-v1.controller.ts
```diff
@@ -36,21 +37,39 @@ export class HandlesControllerV1 {
...
+    if (isValidHandle.isFalse) {
+      throw new BadRequestException('Invalid handle provided!');
+    }
+  }
+
   @Post()
   @HttpCode(HttpStatus.OK)
   @ApiOperation({ summary: 'Request to create a new handle for an account' })
   @ApiOkResponse({ description: 'Handle creation request enqueued', type: TransactionResponse })
+  @ApiBadRequestResponse({ description: 'Invalid handle provided or Provided signature is not valid for the payload!' })
```
mattheworris, line 64: These are updates to the error handling and don't break the API, so I think it is safe to continue with v1.

.github/workflows/e2e-tests.yml
```diff
@@ -63,6 +65,13 @@ jobs:
       matrix:
         service: ${{ fromJson(needs.service-matrix.outputs.services) }}
     steps:
+      - name: Debug Run or Skip Errors
```
mattheworris, line 68: Run of Skip was evaluating to False and the e2e tests did not run.
I added these echoes to debug and then the tests ran. 🤷‍♂️ 
I suggest we leave the echoes for future debugging.

apps/account-api/src/controllers/v2/accounts-v2.controller.ts
```diff
@@ -96,6 +99,16 @@ export class AccountsControllerV2 {
     // Also makes sure it is either a login or a delegation
     const payload = await this.siwfV2Service.getPayload(callbackRequest);
 
+    // Validate claim handle transactions to prevent invalid submissions to the blockchain
+    await Promise.all(
+      payload.payloads.filter(isPayloadClaimHandle).map(async (claimHandle) => {
```
JoeCap08055, line 104: nit/code golf; I find this more readable:
```suggestion
      payload.payloads.filter((payload) => isPayloadClaimHandle(payload)).map(async (claimHandle) => {
```

libs/account-lib/src/utils/blockchain-scanner.service.ts
JoeCap08055: I'm fine leaving it out, but if so it should probably just be removed instead of commented.
Could also change it to `log.debug`--debug level should only be for Gateway developers, not Gateway users.
Either is good.

JoeCap08055: At some point we should probably do some logging cleanup; specifically:
* DEBUG logs should only be relevant for developers working on Gateway itself
    * (understanding, though, that these first clients may need to turn on debug to help us resolve their issues)
* Need a way to turn on VERBOSE logs without also enabling DEBUG
* Consider moving to Pino (as other PL projects) or Winston (very popular) logger

.github/workflows/e2e-tests.yml
JoeCap08055: The tests ran because you've now modified this file, which is included in the "common" set, and therefore triggers "run-all". So the test for "account" modification may still be broken in some way.

PR #681, opened 2024-11-05
fix: use correct test for determine if jobs should run for matrix items

JoeCap08055:
# Problem
Correct suite of matrix CI jobs were not being run based on files included in a PR

# Solution
Use correct logic in the pipeline variable output. Boolean variables cannot be tested alone, apparently, but must be part of a comparison in order to generate a boolean result.

ie,
```yaml
${{ some-var || some-function-call() }}
```
doesn't work, because `some-var` is evaluated as a string. Replace with:
```yaml
# Note, can use true or "true" here, as YAML coerces to a string
${{ some-var == true | | some-function-call() }}
```

Tested locally with (https://github.com/nektos/act)[https://github.com/nektos/act]

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -64,7 +64,7 @@ jobs:
     steps:
       - name: Run or Skip
         id: should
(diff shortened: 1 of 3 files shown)


mattheworris (approved):
- Nice work!
🚀

PR #683, opened 2024-11-09
Developer Docs updates for docker swarm deployment

mattheworris:
# Problem

Docker swarm instructions need to be updated.

# Solution

Created a docker swarm on AWS and updated the directions with more specific instructions and added some debugging commands.

### DIFF
--- a/docs/src/Run/Deployment.md
+++ b/docs/src/Run/Deployment.md
@@ -78,6 +78,9 @@ This guide provides example step-by-step instructions to deploy the Gateway serv
     - TCP/UDP 7946 (communication among nodes).
     - UDP 4789 (overlay network traffic).
     - TCP 2377 (Swarm manager communication).
+  - Allow Gateway service ports (e.g., this will depend on your Swarm mappings, default starts at 30000)
+    - TCP 30000-32767 (Swarm mode routing mesh).
+    - OR specific ports for each service, see `SERVICE_PORT_X` in [docker-compose-swarm.yaml](../../../deployment/swarm/docker-compose.yaml)
 
 #### Step 2: Configure SSH Access
 
@@ -85,6 +88,8 @@ This guide provides example step-by-step instructions to deploy the Gateway serv
 - Note the public IP addresses for each instance.
 
 ### 1.2 Installing Docker Swarm
+
+Login to each EC2 instance (or other cloud instances) using SSH.
 
 #### Step 1: Update Packages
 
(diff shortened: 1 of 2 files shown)


JoeCap08055:
Some notes on IPFS:

Have we tested whether using this configuration inside a cloud provider (or on-prem hosted) allows the Kubo IPFS node to communicate properly with the global IPFS network? Theoretically it should, but may depend on the network environment (for instance, running this locally on a developer's machine behind a residential router usually won't work in that way due to NAT & firewall configuration issues).

We may want to update the docs with some verbage about that, and also the fact that they may also want to consider external IPFS pinning services. Running their own Kubo node could, in effect, be their own pinning service if properly configured, but there are storage & network issues to consider.

Might be the subject of a separate issue...

jeanettedepatie (approved):
A few picky grammar notes.  Otherwise good to go.

docs/src/Run/Deployment.md
```diff
@@ -112,36 +117,86 @@ sudo docker swarm init --advertise-addr <manager-node-ip>
...
 
 #### Step 2: Deploy the Stack
 
+The repo includes an example [docker-compose-swarm.yaml](../../../deployment/swarm/docker-compose.yaml) file for deploying the Gateway services on Docker Swarm.
+Edit the file to set the correct environment variables and service ports.
+Take note of the number of replicas for each service. The default is set to 3.
+
 ```bash
-sudo docker stack deploy -c docker-compose.yaml gateway
+sudo docker stack deploy -c docker-compose-swarm.yaml gateway
```
JoeCap08055, line 158: It might be confusing to refer to this file as `docker-compose-swarm.yaml`, but the example file is `deployment/swarm/docker-compose.yaml`. I'd recommend one of the following:
* update the instructions in step 2 above to specifically copy & rename the example file
* Rename the example file to `deployment/swarm/docker-compose-swarm.yaml`
* Change the subsequent references to simpy refer to `docker-compose.yaml` to match the example filename (my least favoriate)

JoeCap08055 (approved):
🚢 it!

docs/src/Run/README.md
```diff
@@ -8,10 +8,119 @@ Look for the Quick Start guide in the Run Gateway Services section to get starte
...
+```
+
+Each service requires connection to a Redis instance. The `REDIS_URL` environment variable is set to `redis://redis:6379` by default. If you are using a different Redis instance, you can set the `REDIS_URL` environment variable to the appropriate connection string.
+
+Each service also requires a docker network (or equivalent) to connect to any other containers. The default network is set to `gateway_net`. If you are using a different network, you can edit the `networks:` environment variable in the `docker-compose.yaml` to the appropriate network name.
+
+Some services require a connection to an IPFS instance. See the [IPFS Setup Guide](./IPFS.md) for more information.
+
+See the [docker-compose-swarm.yaml](https://github.com/projectlibertylabs/gateway/blob/main/deployment/swarm/docker-compose-swarm.yaml) for examples of redis and ipfs services.
+
```
jeanettedepatie, line 51: This link isn't working for me.  Perhaps it isn't published yet or I don't have access.

jeanettedepatie (approved):

jeanettedepatie:
Just had trouble with one link, but it might just be me.

docs/src/Run/README.md
mattheworris: Yeah, that one isn't merged into main yet, it will be with this PR.

saraswatpuneet (approved):
lgtm

PR #684, opened 2024-11-09
Add SIWF button to the example

wesbiggs:
Updates the example to serve an SVG image for the "Sign in with Frequency" button.
This will give developers a nudge to use the same image in their code.


mattheworris (approved):
Nice update!
🚢 it!

PR #695, opened 2024-11-22
Fix warning overflow

### LINKED ISSUE #693: Warning icon covers text
### Where is the Issue?

https://projectlibertylabs.github.io/gateway/GettingStarted/SSO.html#option-b-dynamic-callback-or-permissions

### What is wrong with it?

The "warning" icon covers up some of the text

image

### What should it be?

I should be able to clearly see the text.

### Any other comments?

Fix should be placed inside `overrides.css`

shannonwells:
# Problem
Same issue in SIWF, fixed the same way; warning icon overflows and blocks warning text.
Fixes #693 

# Solution
Add same changes to overrides.css as in that issue.

## Steps to Verify:
Run `mdbook serve` after building, and go to http://localhost:3000/GettingStarted/SSO.html to check that the issue is fixed.

## Screenshots (optional):
Screenshot 2024-11-22 at 9 43 46 AM

### DIFF
--- a/docs/css/overrides.css
+++ b/docs/css/overrides.css
@@ -289,3 +289,14 @@ hr {
   color: var(--headers);
   margin: 20px 0;
 }
+
+.warning {
+  position: relative;
+}
+
+.warning:before {
+  width: 2.5rem;
+  margin-inline-start: calc(-3rem + 7px);
+  top: -8px;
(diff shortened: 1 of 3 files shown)


wilwade (approved):
🙇

PR #696, opened 2024-12-04
Added spellcheck for markdown files

wesbiggs:
# Problem

tpyos

# Solution

Added `spellchecker-cli`, created its config and dictionary override files, and added to the github build workflow.

I only fixed clear errors.
Others may want to review the dictionary where there are multiple styles in use for the same term (e.g. NodeJS vs. Nodejs vs. nodejs).

Note: `docs/src/CoreConcepts/README.md` is excluded because it contains lorem ipsum text, and that seemed too silly to include in the dictionary.
Otherwise, all `.md` files should be found and checked.

## Steps to Verify:

- `npm run spellcheck` to run locally
- try adding a misspelling in one of the markdown files and see if it picks it up
- try submitting a PR with a misspelling and see if it breaks the build


package.json
```diff
@@ -153,6 +154,7 @@
     "reflect-metadata": "^0.2.0",
     "rxjs": "^7.8.1",
     "short-unique-id": "^5.2.0",
+    "spellchecker-cli": "^7.0.0",
```
wilwade, line 157: We need to move this to a devDep so that it doesn't get included in Gateway production code.

wilwade (changes requested):
Thanks. One small tweak to the package.json and we'll merge it!

package.json
wesbiggs: Doh, my mistake. Fixed.

wilwade (approved):
🥇

PR #699, opened 2025-01-06
correct the name of docker network to match docker compose

demisx:
Fix the typo in the docker network name to match what's being used in the docker compose config files.


mattheworris (approved):
🚀

PR #706, opened 2025-02-10
update docs to correct dependencies on IPFS

demisx:
This PR updates relevant docs to correct the stated dependency on IPFS. The Account and Graph services are not dependent on IPFS, while content services are.


mattheworris (approved):
🚢 it!

PR #707, opened 2025-02-13
remove CHAIN_ENVIRONMENT env var

demisx:
The goal of this PR is to remove `CHAIN_ENVIRONMENT` var from repo as it appears not being used anymore.


JoeCap08055 (approved):
👍🏻

PR #717, opened 2025-02-19
713 content publishing support posting existing at protocol batch content notifications to frequency using a non dsnp schema on frequency

### LINKED ISSUE #713: [content-publishing] support posting existing AT protocol batch content notifications to Frequency using a non-DSNP schema on Frequency
# Description
Currently, `content-publishing-service` allows posting messages for the following schemas:
* `dsnp.broadcast`
* `dsnp.reply`
* `dsnp.reaction`
* `dsnp.tombstone`
via specific endpoints in the API controller, `POST /broadcast`, `POST /reply`, `POST /reaction`, and `DELETE`

`content-publishing-service` receives requests to post individual messages for the indicated schemas, and handles creating batch files which then get posted to IPFS, and the resulting CIDv1 content hash is posted to Frequency, creating a Message announcement.

To support Bluesky's AT Protocol, which already has batched content hosted on AWS S3, we need to create appropriate endpoints and backend support for announcing AT batched content directly to Frequency, without the need to create the batches ourselves.

## Acceptance Criteria
- [ ] New endpoint that accepts the following and enqueues for posting on-chain message to the chain:
    - Schema ID
    - Arbitrary payload bytes
    - On behalf of MSA ID (optional)
-  Validations:
    - [ ] Schema payload location: `OnChain`
    - [ ] `onBehalfOf` MSA ID (if provided) has active delegation for schema to current provider
- [ ] Worker submits message to the chain via the `add_onchain_message` extrinsic

Related questions:
#715
#716

JoeCap08055:
# Description
Create a new V2 content endpoint in `content-publishing`, `POST /v2/content/:schemaId/onChainContent` that allows posting message payloads to schemas with an `OnChain` payload location.

* Creates new endpoint & processing workflow for OnChain messages
* Refactor/removal of a duplicate type def
* Rename of some methods in `BlockchainRpcQueryService` for better clarity

Closes #713


apps/content-publishing-api/src/controllers/v2/content.controller.v2.ts
```diff
@@ -0,0 +1,54 @@
...
+@ApiTags('v2/content')
+export class ContentControllerV2 {
+  constructor(
+    private apiService: ApiService,
+    private readonly blockchainService: BlockchainRpcQueryService,
+    // eslint-disable-next-line no-empty-function
+  ) {}
+
+  // eslint-disable-next-line class-methods-use-this
+  @Post(':schemaId/onChainContent')
```
JoeCap08055, line 28: Happy to entertain other names for this endpoint

apps/content-publishing-api/src/api.service.ts
```diff
@@ -33,10 +34,36 @@ export class ApiService {
...
     this.logger = new Logger(this.constructor.name);
   }
 
+  async enqueueContent(
+    schemaId: number,
+    contentDto: OnChainContentDto,
+    onBehalfOf: string,
+  ): Promise<AnnouncementResponseDto> {
+    const jobData: IPublisherJob = {
+      id: Date.now().toString(), // timestamp will get overwritten, but guarantees id-uniqueness
```
aramikm, line 48: so the whole reason behind calculating JobId was to ensure duplications are detected in queue level to some degree. If you add this here it would cause every request to be unique which would cause duplications for the same content

apps/content-publishing-api/src/api.service.ts
```diff
@@ -33,10 +34,36 @@ export class ApiService {
...
+      data: {
+        onBehalfOf,
+        payload: contentDto.payload,
+      },
+    };
+    jobData.id = this.calculateJobId(jobData);
+    const job = await this.publishQueue.add(`OnChain content job - ${jobData.id}`, jobData, {
+      jobId: jobData.id,
+      removeOnComplete: 1000,
+      attempts: 3,
```
aramikm, line 59: I don't have strong feelings about this but having attempt in here might be a little be tricky. Since this is in the api we probably want it to fail fast. Because the caller can always retry the call itself

aramikm (approved):
Read the code. Overall looks good. A few comments

apps/content-publishing-api/src/controllers/v2/content.controller.v2.ts
mattheworris: Maybe `anchoredContent`?

apps/content-publishing-api/src/api.service.ts
JoeCap08055: This was the existing standard in the app. I'll check the failure conditions, but we do sanity checks before submitting to the chain, so mostly a failure would be for network or Capacity reasons, in which case I think it's reasonable to retry (Capacity exhaustion would pause the queue, anyway).

I'll take a pass through the code & see if I can trap any specific chain errors that we would want to trigger a permanent failure, if we're not already doing that.

JoeCap08055: Okay, so the only _extrinsic_ failure that would cause a retry is `TooManyMessagesInBlock`, which seems reasonable. Any other _extrinsic_ failure explicitly skips the retry.

Other failures, like network errors, etc, _do_ cause a retry, which also seems reasonable.

jeetumaker (approved):

enddynayn (approved):
👍  great!

PR #720, opened 2025-02-21
feat: add support for publishing batch announcements for non-DSNP schemas

### LINKED ISSUE #719: [content-publishing] Enable publishing of pre-built batch files to arbitrary IPFS schemas
# Description
`content-publishing` should allow publishing client-built Parquet batch files to arbitray schemas with payload location `IPFS`

# Acceptance Criteria
- [ ] `asset` controller modified to allow upload of `application/vnd.apache.parquet` assets
- [ ] `content` controller endpoint added:
    - `/v2/content/batchAnnouncement` endpoint takes `schemaId` & `referenceId` and enqueues a job to post the batch file to the chain
- [ ] "worker" app adds processing to take enqueued job and post to the chain

JoeCap08055:
# Description
- asset controller modified to allow upload of application/vnd.apache.parquet assets
- content controller endpoint added:
    - /v2/content/batchAnnouncement endpoint takes schemaId & referenceId and enqueues a job to post the batch file to the chain
    - "worker" app adds processing to take enqueued job and post to the chain

Closes #719


libs/utils/src/decorators/is-dsnp-content-hash.decorator.ts
JoeCap08055, line 1: I tested this with valid CIDv1 and it failed. Apparently it was only specified on Tombstone & Update messages, which are not currently supported by the Social App Template; hence we never tripped over it.

Re-worked the logic according to the official CID decoding algorithm.

apps/content-publishing-worker/src/batch_announcer/batch.announcer.ts
```diff
@@ -69,6 +69,17 @@ export class BatchAnnouncer {
     return { id: batchId, schemaId, data: { cid, payloadLength: size } };
   }
 
+  public async announceExistingBatch(batch: IBatchFile): Promise<IPublisherJob> {
```
aramikm, line 72: nit: I think this is mostly fetching it instead of announcing it so the name is a bit confusing.

libs/utils/src/decorators/is-dsnp-content-hash.decorator.ts
```diff
@@ -12,18 +13,36 @@ export function validateContentHash(contentHash: unknown): boolean {
   }
 
   try {
+    if (contentHash.length === 46 && contentHash.startsWith('Qm')) {
```
aramikm, line 16: From my understanding dsnp multihash is different from IPFS CID and I don't think we were supporting CID directly as dsnp content hash

aramikm (approved):
Added a comment about CID and dsnp content hash but overall looks good

libs/utils/src/decorators/is-dsnp-content-hash.decorator.ts
aramikm: Based on my understanding of the spec It's not supposed to pass because they are different values

mattheworris (approved):
- Read through the code.
Looks good, 🚢 it!

libs/utils/src/decorators/is-dsnp-content-hash.decorator.ts
JoeCap08055: There's a bit of a disconnect. The DSNP spec only specifies that it must be a valid base32 multibase encoding with one of the DSNP-supported multihash hashing algorithms (Blake3 or SHA2-256):
https://spec.dsnp.org/DSNP/Identifiers.html#dsnp-content-hash

However, the DSNP spec does not specify a storage location. The _DSNP over Frequency_ spec, however, specifies that off-chain content is specifically hosted on IPFS and that announcements of off-chain messages must be validated as CIDv1:
https://spec.dsnp.org/Frequency/Validation.html

I do see one problem, however: the Frequency `messages` pallet validates as a CIDv1, but does not validate agains the DSNP-spec; ie, CIDv1 only requires a valid multibase format encoding of a valid multihash-encoded hash, while DSNP narrows that further to a specific multibase encoding (`base32`) and one of 2 specific multihash hashing algorithms.

Based on that, I think this code in Gateway satisfies _both_ DSNP _and_ Frequency/CIDv1 specs, while the code in the Frequency chain itself might technically allow a non-DSNP-compliant CIDv1 to pass.

apps/content-publishing-worker/src/batch_announcer/batch.announcer.ts
JoeCap08055: Just following the existing nomenclature here; the existing `announce` function is _creating_ a batch file & pinning to IPFS, then creating the payload for the on-chain announcement.

This new function skips creating the batch & pinning it, since that's already been done, and simply creates the announcement payload.

I agree that the naming doesn't strictly describe what's being done, but didn't want to get into a whole nomenclature change in this PR.

PR #743, opened 2025-03-14
fix: minor miscellaneous updates

mattheworris:
# Problem

`content-publishing` module configuration was missing imports, so the worker would not run. There are other miscellaneous bug fixes included.

# Solution

Corrected miscellaneous bugs.


JoeCap08055 (approved):
Looks good--I'm curious how the pipeline ran successfully on the PR that implemented the new `HTTP_RESPONSE_TIMEOUT_MS` 🤔

PR #752, opened 2025-03-21
fix: do not queue multiple jobs when same SIWF payload is posted multiple times

### LINKED ISSUE #674: [bug] Handle Duplicate POSTS to /v2/accounts/siwf
Problem: Right now if someone double posts to `/v2/accounts/siwf` things go bad.

Currently:
- Adds the action to the queue
- Queue action fails

Expected:
- Idempotent
- Can post as many times as is valid
- Only results in one queued action
- IF the queued action is already complete, should return the MSA id and other information as per normal

Question: What about replays?
- That should be handled by the service handling user session creation

JoeCap08055:
# Problem
If the same authorizationCode was posted to the `/v2/accounts/siwf` endpoint, multiple queue tasks would be created for (essentially) the same payload (at least, the same sequence of extrinsics).

# Root Cause
The work queue in BullMQ is intended to be unique on `jobId`. The `jobId` was being created from a hash of the payload retrieved from SIWF (Frequency Access). However, when a payload is retrieved from SIWF using an authorization code, that payload is *SIGNED ON DEMAND*--meaning the signed extrinsics within the payload differ; different contents => different hash, different job ID, hence multiple jobs.

# Solution
If an authorization code is being used to retrieve a payload, use the authorization code as the job ID instead of the retrieved payload hash

Note, it's still possible that the same operations could be submitted by an external agent who handles their own signing/payload creation, or handles fetching the payload from SIWF themselves and submits the actual payload to Gateway instead of the authorization code. This solution does not prevent that scenario from causing multiple jobs to be queued for the same sequence of user creation transactions. (Tracking as #753)

Closes #674


apps/account-api/src/services/siwfV2.service.ts
```diff
@@ -177,12 +181,9 @@ export class SiwfV2Service {
...
 
-    // Try to look up the MSA id, if there is no createSponsoredAccountWithDelegation request
-    if (payload.payloads.every((x) => x.endpoint?.extrinsic !== 'createSponsoredAccountWithDelegation')) {
-      // Get the MSA Id from the chain
-      const msaId = await this.blockchainService.publicKeyToMsaId(response.controlKey);
-      if (msaId) response.msaId = msaId;
-    }
+    // Get the MSA Id from the chain, if it exists
+    const msaId = await this.blockchainService.publicKeyToMsaId(response.controlKey);
+    if (msaId) response.msaId = msaId;
```
JoeCap08055, line 186: Because it's possible the same payload could be submitted multiple times, we should check for the MSA anyway. (It doesn't hurt; only costs an RPC query to the chain)

libs/account-lib/src/services/enqueue-request.service.ts
```diff
@@ -19,30 +18,49 @@ export class EnqueueService {
     this.logger = new Logger(this.constructor.name);
   }
 
-  private calculateJobId<RequestType>(jobWithoutId: RequestType): string {
+  private static calculateJobId<RequestType>(jobWithoutId: RequestType): string {
```
JoeCap08055, line 21: Changed to static method so we don't need to suppress the linter

libs/types/src/account-webhook/types.gen.ts
JoeCap08055, line 1: Generated file; the tool was updated and now adds these types in; can't seem to turn them off, but no harm.

libs/types/src/dtos/account/transaction.response.dto.ts
JoeCap08055, line 1: Removed all of the `class-validator` decorators from this file, since they're response types and therefore validation is not invoked.
Normally, responses could just be interfaces, not classes, but we need the Swagger decorators to generate the docs properly, and you can't attach decorators to interfaces.

libs/types/src/dtos/account/wallet.v2.login.response.dto.ts
JoeCap08055, line 1: Same as above; no need for class-validator decorations on response types.

aramikm: nit: Probably none of these are used in any requests but it wouldn't hurt to double check specially the inner types.

openapi-ts.config.ts
```diff
@@ -2,16 +2,25 @@
...
+      dates: true,
+      name: '@hey-api/transformers',
+    },
+    {
+      identifierCase: 'preserve',
+      enums: 'typescript',
+      enumsCase: 'preserve',
+      name: '@hey-api/typescript',
+    },
+  ],
```
aramikm, line 23: q: Any reason why this is changed?

aramikm (approved):
Overall looks great! some comments and nits.

libs/account-lib/src/services/enqueue-request.service.ts
```diff
@@ -19,30 +18,49 @@ export class EnqueueService {
...
+    // For paylaods submitted directly, we can de-dupe IFF the exact same payload
+    // is submitted multiple times; however, it is possible that the same set of extrinsic
+    // calls could be submitted multiple times, but with different signatures. In that case
+    // the only way to de-dupe would be to decode the extrinsics in the payload and de-dupe
+    // based on examination of the public key and other parameters of the extrinsics. While
+    // technically possible, deemed overkill for now.
+    const authorizationCode =
+      'authorizationCode' in request && typeof request.authorizationCode === 'string'
+        ? request.authorizationCode
+        : undefined;
```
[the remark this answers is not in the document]
JoeCap08055: In the end, this is really a very specific case. SIWFv2 is the only case where the submitted payload could consist of _either_ a full payload _or_ a code used to retrieve a payload that could differ due to signatures. So I think trying to develop a generic facility here would be overkill.

libs/types/src/dtos/account/wallet.v2.login.response.dto.ts
JoeCap08055: Confirmed.

PR #754, opened 2025-03-21
Fix type generation from OpenAPI spec

JoeCap08055:
# Problem
A recent change to an OpenAPI spec (and possibly some changes to the tool used to generate Typescript types from OpenAPI) caused `content-watcher` to fail to build after types were regenerated.

# Solution
Since no config options were available in the tool, added a post-processing script to the type generation to manually correct the generated objects.


wilwade (approved):
Looks good to me!

- Pulled and locally generated the HTML and viewed it.
- Pulled and locally generated the JSON and viewed it.

Issue #750, opened 2025-03-20
Performance Bug: Account Service Worker slow to dequeue


wilwade:
Problem: Account Service Worker currently only handles about 10-20 queued signup jobs per block.
Expected: It can handle up to 200 per block

Notes:
- @wilwade has a hacked together load test for this
- Watch out for capacity limit issues
- concurrency: 200 is likely fine, but we should check to see about nonce and capacity issues and how to fix them

Quick Patch that cranks it up to 200 with no safeties:

```
diff --git forkSrcPrefix/apps/account-worker/src/transaction_publisher/publisher.service.ts forkDstPrefix/apps/account-worker/src/transaction_publisher/publisher.service.ts
index 89d913c31ff24befff7247de47c42b44184769a9..7358e9cb3c4891fb01dda5ce2b2733d694edb992 100644
--- forkSrcPrefix/apps/account-worker/src/transaction_publisher/publisher.service.ts
+++ forkDstPrefix/apps/account-worker/src/transaction_publisher/publisher.service.ts
@@ -33,7 +33,9 @@ const CAPACITY_EPOCH_TIMEOUT_NAME = 'capacity_check';
  * Service responsible for publishing account updates.
  */
 @Injectable()
-@Processor(QueueConstants.TRANSACTION_PUBLISH_QUEUE)
+@Processor(QueueConstants.TRANSACTION_PUBLISH_QUEUE, {
+  concurrency: 200,
+})
 export class TransactionPublisherService extends BaseConsumer implements OnApplicationShutdown {
   public async onApplicationBootstrap() {
     await this.capacityCheckerService.checkForSufficientCapacity();
```

PR #758, opened 2025-03-25
feat: eliminate some chain RPC calls in favor of cached values

JoeCap08055:
This PR attempts to eliminate a number of RPC calls to the chain from the critical path of submitting a transaction. By caching certain values retrieved from the chain, we both increase job throughput in the Gateway services, and reduce load on the chain RPC node.

- Makes all queue workers' concurrencies configurable via env `<queueName>_QUEUE_WORKER_CONCURRENCY`
- Manually specify block hash & era for transaction signing and submission; eliminates anywhere from 3-7 RPC calls to the chain
- Keep current capacity usage and latest nonce updated in Redis cache on a defined interval. This takes it out of the critical path of submitting a transaction. (It's OK not to get the absolute latest from the chain each time)
- If we detect an error submitting a transaction, un-reserve the none key that we submitted

Related issue: #750

### DIFF
--- a/apps/account-worker/src/transaction_publisher/publisher.service.ts
+++ b/apps/account-worker/src/transaction_publisher/publisher.service.ts
@@ -5,7 +5,6 @@ import { DelayedError, Job, Queue } from 'bullmq';
 import Redis from 'ioredis';
 import { SubmittableExtrinsic } from '@polkadot/api-base/types';
 import { IMethod, ISubmittableResult, Signer, SignerResult } from '@polkadot/types/types';
-import { MILLISECONDS_PER_SECOND } from 'time-constants';
 import { SchedulerRegistry } from '@nestjs/schedule';
 import { BlockchainService } from '#blockchain/blockchain.service';
 import { ICapacityInfo } from '#blockchain/types';
@@ -27,7 +26,6 @@ import { getSignerForRawSignature } from '#utils/common/signature.util';
 import { TXN_WATCH_LIST_KEY } from '#types/constants';
 import workerConfig, { IAccountWorkerConfig } from '#account-worker/worker.config';
 
-export const SECONDS_PER_BLOCK = 6;
 const CAPACITY_EPOCH_TIMEOUT_NAME = 'capacity_check';
 
 /**
@@ -215,7 +213,7 @@ export class TransactionPublisherService extends BaseConsumer implements OnAppli
   public async handleCapacityExhausted(capacityInfo: ICapacityInfo) {
     await this.transactionPublishQueue.pause();
     const blocksRemaining = capacityInfo.nextEpochStart - capacityInfo.currentBlockNumber;
-    const epochTimeout = blocksRemaining * SECONDS_PER_BLOCK * MILLISECONDS_PER_SECOND;
+    const epochTimeout = blocksRemaining * (this.blockchainService.blockTimeMs || 6000);
     // Avoid spamming the log
     if (!(await this.transactionPublishQueue.isPaused())) {
       this.logger.warn(
@@ -236,7 +234,7 @@ export class TransactionPublisherService extends BaseConsumer implements OnAppli
       );
     } catch (err) {
       // Handle any errors
-      console.error(err);
+      this.logger.error(err);
     }
   }
 
--- a/apps/account-worker/src/transaction_publisher/publisher.service.ts
+++ b/apps/account-worker/src/transaction_publisher/publisher.service.ts
@@ -1,6 +1,6 @@
 import { InjectRedis } from '@songkeys/nestjs-redis';
 import { InjectQueue, Processor } from '@nestjs/bullmq';
-import { Injectable, OnApplicationShutdown } from '@nestjs/common';
+import { Inject, Injectable, OnApplicationBootstrap, OnApplicationShutdown } from '@nestjs/common';
 import { DelayedError, Job, Queue } from 'bullmq';
 import Redis from 'ioredis';
 import { SubmittableExtrinsic } from '@polkadot/api-base/types';
@@ -25,6 +25,7 @@ import { Vec } from '@polkadot/types';
 import { Call } from '@polkadot/types/interfaces';
 import { getSignerForRawSignature } from '#utils/common/signature.util';
 import { TXN_WATCH_LIST_KEY } from '#types/constants';
+import workerConfig, { IAccountWorkerConfig } from '#account-worker/worker.config';
 
 export const SECONDS_PER_BLOCK = 6;
 const CAPACITY_EPOCH_TIMEOUT_NAME = 'capacity_check';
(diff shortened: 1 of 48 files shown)


libs/blockchain/src/blockchain.service.ts
```diff
@@ -159,14 +159,19 @@ export class BlockchainService extends BlockchainRpcQueryService implements OnAp
...
+    try {
+      this.logger.debug(`Capacity Wrapped Extrinsic: ${JSON.stringify(extrinsic.toHuman())}, nonce: ${nonce}`);
+      const txHash = await extrinsic.signAndSend(keys, { nonce });
+      if (!txHash) {
+        throw new Error('Tx hash is undefined');
+      }
+      this.logger.debug(`Tx hash: ${txHash}`);
+      return [extrinsic, txHash.toHex()];
+    } catch (err: any) {
+      this.unreserveNonce(nonce);
```
aramikm, line 172: Should this be
```suggestion
      await this.unreserveNonce(nonce);
```

libs/blockchain/src/blockchain.service.ts
```diff
@@ -178,30 +183,44 @@ export class BlockchainService extends BlockchainRpcQueryService implements OnAp
...
+    try {
+      this.logger.debug(`Capacity Wrapped Extrinsic: ${JSON.stringify(extrinsic.toHuman())}, nonce: ${nonce}`);
+      const txHash = await extrinsic.signAndSend(keys, { nonce });
+      if (!txHash) {
+        throw new Error('Tx hash is undefined');
+      }
+      this.logger.debug(`Tx hash: ${txHash.toString()}`);
+      return [extrinsic, txHash.toHex()];
+    } catch (err: any) {
+      this.unreserveNonce(nonce);
```
aramikm, line 196: This might be intentional but It might make sense to wait for it. It might prevent some funky edge cases
```suggestion
      await this.unreserveNonce(nonce);
```

libs/blockchain/src/blockchain.service.ts
```diff
@@ -178,30 +183,44 @@ export class BlockchainService extends BlockchainRpcQueryService implements OnAp
...
-  public async getNextNonce(): Promise<number> {
-    const nonce = await this.getNonce(this.accountId);
-    const keys = getNextPossibleNonceKeys(nonce);
+  public async reserveNextNonce(): Promise<number> {
+    const currentNonceKey = getNonceKey(this.accountId, 'current');
+    let nonce: string | number = await this.nonceRedis.get(currentNonceKey);
+    if (!nonce) {
+      nonce = await this.getNonce(this.accountId);
+      await this.nonceRedis.setex(currentNonceKey, NONCE_KEY_EXPIRE_SECONDS / 2, nonce);
+    }
```
aramikm, line 207: nit: this can be refactored into it's own small function

libs/blockchain/src/capacity-checker.service.ts
```diff
@@ -11,37 +11,65 @@
...
     private readonly blockchainService: BlockchainRpcQueryService,
-    private readonly configService: ConfigService,
+    @Inject(blockchainConfig.KEY) private readonly config: IBlockchainConfig,
     @InjectRedis() private readonly redis: Redis,
     private readonly eventEmitter: EventEmitter2,
-    // eslint-disable-next-line no-empty-function
   ) {
     this.logger = new Logger();
+
+    // NOTE: Why we do this on an interval instead of using an expiring cache key & fetching inside checkForSufficientCapacity:
```
aramikm, line 58: I might be wrong but I don't think setting the cache on Redis and querying it for every request would be that much of a overhead. Calling chain RPC is more expensive but I assume Redis calls should be really fast. Do we have any measurement to support that this is necessary?

JoeCap08055: The idea here is that, in the case where the key has expired & the capacity needs to be fetched from the chain, we don't want that RPC call in the critical path of submitting a transaction. Better to have it update in a separate async task. (Yes, I realize that given the single-threaded nature of Node this may be a fool's errand, but at least I feel better about it 😅 )

aramikm (approved):
Looks good! I think we might also want to document the block header related optimizations somewhere maybe along with nonce stuff so people don't get confused.

libs/blockchain/src/blockchain.service.ts
```diff
@@ -150,59 +198,126 @@ export class BlockchainService extends BlockchainRpcQueryService implements OnAp
...
+    await this.nonceRedis.del(getNonceKey(this.accountId, `${nonce}`));
+  }
+
   public createTxFromEncoded(encodedTx: any): SubmittableExtrinsic<'promise', ISubmittableResult> {
     return this.api.tx(encodedTx);
   }
+
+  /**
+   * Gets the block hash and number of the latest block for signing. We cache this info & update asynchronously. This
+   * elminates unnecessary RPC calls to get the latest block info, since it's okay if we're a little behind in the block number/hash we
```
mattheworris, line 294:
```suggestion
   * eliminates unnecessary RPC calls to get the latest block info, since it's okay if we're a little behind in the block number/hash we
```

libs/blockchain/src/blockchain.service.ts
```diff
@@ -150,59 +198,126 @@ export class BlockchainService extends BlockchainRpcQueryService implements OnAp
...
+    ]);
+
+    const latestHeader = JSON.parse(latestHeaderStr) as IHeaderInfo;
+    const finalizedHeader = JSON.parse(finalizedHeaderStr) as IHeaderInfo;
+
+    return latestHeader.number - finalizedHeader.number > MAX_FINALITY_LAG.toNumber() ? finalizedHeader : latestHeader;
+  }
+
+  /**
+   * Gets get default mortality period for transactions.
```
mattheworris, line 311:
```suggestion
   * Gets default mortality period for transactions.
```

apps/account-worker/src/transaction_publisher/publisher.service.ts
```diff
@@ -196,7 +196,11 @@ export class TransactionPublisherService extends BaseConsumer implements OnAppli
       if (!txHash) throw new Error('Tx hash is undefined');
 
       this.logger.debug(`Tx hash: ${txHash}`);
-      return [submittableExtrinsic, txHash];
+
+      // The caller needs to know the block number (or approximate block number) of the block hash used when the transaction was signed.
+      // But because these "proxy" transactions are signed externally, we don't know that. All we can do is use the current block number as a best guess.
```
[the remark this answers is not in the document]
wilwade: I believe all it does is reduces the mortality window. Off by too much and RPC error due to bad mortality.

wilwade (commented):
Noting that I have pulled and run a fair number of load tests against this PR and it has performed well.

mattheworris (approved):
- Read through changes

🚢 it!

apps/account-worker/src/transaction_publisher/publisher.service.ts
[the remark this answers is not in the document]
JoeCap08055: Correct.

libs/blockchain/src/capacity-checker.service.ts
JoeCap08055: There's still one capacity check in the critical path. The cached check just checks to see if we've exceeded our configured limits.

There's a per-transaction check that makes sure we actually have enough capacity for this specific extrinsic, that actually calls the chain. We want to get this out of the critical path as well, and then catch any out-of-capacity errors and re-queue, but this PR is big enough already. Tracking in #760

PR #761, opened 2025-03-27
feat: implement dynamic log level configuration across applications

### LINKED ISSUE #675: Blockchain Scanner is logging a little too profusely
### Steps to reproduce

1. Run the gateway services.
2. Observe the following logs for `account-worker` make up most of the log

```text
[Nest] 31312  - 11/04/2024, 3:22:44 PM VERBOSE [TxnNotifierService] Starting scan from block #1961501
[Nest] 31312  - 11/04/2024, 3:22:45 PM   DEBUG [TxnNotifierService] Latest finalized block (1961501) encountered
```

### Expected behavior
- [x] `account-worker` blockchain scan logs are not too profuse
- [x] `content-publishing-worker` blockchain scan logs are not too profuse
- [ ] #762
- [x] `graph-worker` blockchain scan logs are not too profuse

Blockchain scanner should not log unless it's found something in the scan.

mattheworris:
# Problem

See #675

# Solution

This PR enhances log level control by refining how the DEBUG environment variable is used and introducing a new VERBOSE_LOGGING variable for more granular configuration.

Changes

- The `DEBUG=true` environment variable now only enables the debug log level.
- To enable both `debug` and `verbose` log levels, set `DEBUG=verbose`.
- `DEBUG` is defined in the docker-compose.yaml file under the common environment section, enabling debug logging across all gateway services by default.

New Feature

- Introduces the `VERBOSE_LOGGING` environment variable to enable `verbose` logging on a per-service basis.

- It is not defined in the shared `docker-compose.yaml`.
- It must be set in the individual `.env.[service]` file.
- For deployments, it can optionally be added to `docker-compose.yaml` as needed.

This setup makes it easier to enable detailed logs across all services or selectively on a per-service basis.

Closes #675 

## Steps to Verify:

1. Use `start.sh` to configure running with dev containers (NOT published containers).
2. Verify debug logging by default.
3. Modify either `.env.[service]` to include `DEBUG=verbose` or `VERBOSE_LOGGING=true`.
4. Use `stop.sh` and then restart the services using `start.sh` so that environment changes will take effect.
5. Verify verbose logging.

### DIFF
--- a/libs/content-publishing-lib/src/utils/blockchain-scanner.service.ts
+++ b/libs/content-publishing-lib/src/utils/blockchain-scanner.service.ts
@@ -113,7 +113,6 @@ export abstract class BlockchainScannerService {
       }
     } catch (e) {
       if (e instanceof EndOfChainError) {
-        this.logger.verbose(e.message);
         return;
       }
 
--- a/libs/content-publishing-lib/src/utils/blockchain-scanner.service.ts
+++ b/libs/content-publishing-lib/src/utils/blockchain-scanner.service.ts
@@ -113,7 +113,7 @@ export abstract class BlockchainScannerService {
       }
     } catch (e) {
       if (e instanceof EndOfChainError) {
-        this.logger.debug(e.message);
+        this.logger.verbose(e.message);
         return;
       }
 
--- a/apps/account-api/src/main.ts
+++ b/apps/account-api/src/main.ts
@@ -7,6 +7,7 @@ import { TimeoutInterceptor } from '#utils/interceptors/timeout.interceptor';
 import { NestExpressApplication } from '@nestjs/platform-express';
 import apiConfig, { IAccountApiConfig } from './api.config';
 import { generateSwaggerDoc, initializeSwaggerUI, writeOpenApiFile } from '#openapi/openapi';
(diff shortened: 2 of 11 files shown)


libs/content-publishing-lib/src/utils/blockchain-scanner.service.ts
```diff
@@ -113,7 +113,7 @@ export abstract class BlockchainScannerService {
       }
     } catch (e) {
       if (e instanceof EndOfChainError) {
-        this.logger.debug(e.message);
+        this.logger.verbose(e.message);
```
JoeCap08055, line 116: Honestly, I think this message is SOOO spammy, we should just get rid of it. Same with the one in the graph worker.
...or... wrap it in some logic so it only logs the first time it's encountered (ie, we might want to see a log after we first start up the scan when we finally get to the end of the chain, but from that point on, we don't need to see it repeatedly)

mattheworris: Yeah, it was already removed from account-service, will remove.

JoeCap08055 (approved):

PR #765, opened 2025-03-28
feat: enhance content-watcher logging

### LINKED ISSUE #762: `content-watcher` blockchain scan logs are not too profuse
## Description

Content-watcher logs are very profuse and impede debugging. Some logic is needed to thin them out.

## Acceptance Criteria

- [x] Content-watcher logs are much less profuse, making debugging easier.
- [x] Logs that cannot just be shoved into verbose should be sampled, or printed in intervals.

mattheworris:
## Description

Content-watcher logs are very profuse and impede debugging. Some logic is needed to thin them out.

## Acceptance Criteria

- [x] Content-watcher logs are much less profuse, making debugging easier.
- [x] Logs that cannot just be shoved into verbose should be sampled, or printed in intervals.

Closes #762

# Solution

- Moved some logs to verbose
- Implemented throttling on some logs
- Miscellaneous fixes for docker files

### DIFF
--- a/libs/content-watcher-lib/src/scanner/scanner.service.ts
+++ b/libs/content-watcher-lib/src/scanner/scanner.service.ts
@@ -141,7 +141,7 @@ export class ScannerService implements OnApplicationBootstrap, OnApplicationShut
 
         const messages = await this.chainEventProcessor.getMessagesInBlock(currentBlockNumber, eventsToWatch);
         if (messages.length > 0) {
-          this.logger.debug(`Found ${messages.length} messages to process`);
+          this.logger.verbose(`Found ${messages.length} messages to process`);
         }
         await ChainEventProcessorService.queueIPFSJobs(messages, this.ipfsQueue);
         await this.saveProgress(currentBlockNumber);


deployment/swarm/docker-compose-swarm.yaml
```diff
@@ -45,7 +45,7 @@ x-graph-service-env: &graph-service-env
 x-account-service-env: &account-service-env
   BLOCKCHAIN_SCAN_INTERVAL_SECONDS: 1
   TRUST_UNFINALIZED_BLOCKS: 'true'
-  WEBHOOK_BASE_URL: 'http://your-app.com:${ACCOUNT_WEBHOOK_PORT:-3001}/webhooks/account-service'
+  WEBHOOK_BASE_URL: '${WEBHOOK_BASE_URL:-http://your-app.com:3001}/webhooks/account-service}'
```
mattheworris, line 48: Took out the `ACCOUNT_WEBHOOK_PORT` so only `WEBHOOK_BASE_URL` needs to be set, added some notes in the template to explain how the port is linked to the webhook.

apps/content-watcher/src/ipfs/ipfs.processor.ts
```diff
@@ -43,7 +43,7 @@ export class IPFSContentProcessor extends BaseConsumer implements OnApplicationB
 
   async process(job: Job<IIPFSJob, any, string>): Promise<any> {
     try {
-      this.logger.log(`IPFS Processing job ${job.id}`);
+      this.logger.verbose(`IPFS Processing job ${job.id}`);
```
mattheworris, line 46: I considered removing these altogether, they are very verbose on the current testnet, but hidden in `verbose` may be good enough.

docker-compose-published.yaml
```diff
@@ -139,7 +139,7 @@ services:
     networks:
       - gateway-net
     profiles:
-      - content_publishing
+      - content-publishing
```
mattheworris, line 142: Making profile names consistent.

docker-compose.yaml
```diff
@@ -17,7 +17,7 @@ x-common-environment: &common-environment
   IPFS_BASIC_AUTH_USER: ${IPFS_BASIC_AUTH_USER:-""}
   IPFS_BASIC_AUTH_SECRET: ${IPFS_BASIC_AUTH_SECRET:-""}
   QUEUE_HIGH_WATER: 1000
-  DEBUG: true
+  DEBUG: ${DEBUG:-true}
```
mattheworris, line 20: This doesn't seem to be working with `DEBUG=verbose`, but could be something to do with running with dev containers.

libs/content-watcher-lib/src/scanner/scanner.service.ts
```diff
@@ -111,21 +114,27 @@ export class ScannerService implements OnApplicationBootstrap, OnApplicationShut
...
         const queueSize = await this.ipfsQueue.count();
         if (queueSize > this.config.queueHighWater) {
-          this.logger.log('Queue soft limit reached; pausing scan until next interval');
+          const now = Date.now();
+          if (now - this.lastQueueLimitLogTime > this.LOG_THROTTLE_MS) {
+            this.logger.log(
+              `Queue soft limit reached (${queueSize}/${this.config.queueHighWater}); pausing scan until next interval`,
+            );
+            this.lastQueueLimitLogTime = now;
+          }
```
JoeCap08055, line 123: Another way to do this would be to introduce a class variable `isHighWater`. Set true when soft limit reached, false if queue has drained enough.

This way we can log when we reach high water, but we don't have to keep logging while we're still in that state. But if we drain & reach high water again, we'll still get the log.

It can be useful in the log to know when we transition between the states, but not keep logging while we're in the state.

libs/content-watcher-lib/src/scanner/scanner.service.ts
```diff
@@ -111,21 +114,27 @@ export class ScannerService implements OnApplicationBootstrap, OnApplicationShut
...
           break;
         }
 
         const currentBlockNumber = await this.getNextBlockNumber();
         const currentBlockHash = await this.blockchainService.getBlockHash(currentBlockNumber);
         if (currentBlockHash.isEmpty) {
-          this.logger.debug(
+          this.logger.verbose(
```
JoeCap08055, line 130: We can probably just get rid of this one altogether.

libs/content-watcher-lib/src/scanner/scanner.service.ts
```diff
@@ -111,21 +114,27 @@ export class ScannerService implements OnApplicationBootstrap, OnApplicationShut
...
             `No new blocks to scan @ block number ${currentBlockNumber}; pausing scan until next interval`,
           );
           break;
         }
 
         if (first) {
-          this.logger.log(`Starting scan @ block # ${currentBlockNumber} (${currentBlockHash})`);
+          this.logger.verbose(`Starting scan @ block # ${currentBlockNumber} (${currentBlockHash})`);
```
JoeCap08055, line 137: Same with this; I vote to get rid of it.

libs/content-watcher-lib/src/scanner/scanner.service.ts
mattheworris: Will remove.

libs/content-watcher-lib/src/scanner/scanner.service.ts
mattheworris: Will remove.

JoeCap08055 (approved):

PR #766, opened 2025-03-28
feat: allow task retry on nonce conflict error

### LINKED ISSUE #751: Bug: Nonce Errors don't always retry
At least for the Account Service `TRANSACTION_PUBLISH_QUEUE`, when we get a nonce collision error, the transaction does not attempt a retry:

```
[Nest] 454  - 03/21/2025, 1:10:43 PM   ERROR [TransactionPublisherService] Unknown error encountered: 
[Nest] 454  - 03/21/2025, 1:10:43 PM   ERROR [TransactionPublisherService] RpcError: 1014: Priority is too low: (196 vs 196): The transaction has too low priority to replace another transaction already in the pool.
RpcError: 1014: Priority is too low: (196 vs 196): The transaction has too low priority to replace another transaction already in the pool.
```

This sort of error, while it shouldn't happen once #750 is correctly fixed, should still result in a clean retry

While this is specifically for the account service, I think all errors of this type are clear retry candidates

JoeCap08055:
This PR catches RPC errors on chain extrinsic submission. If the error is due to a nonce conflict, throw a "DelayedError" which will trigger the initiating queue job to go back into the queue for retry (bypassing configured queue retry limits)

Closes #751

### DIFF
--- a/apps/account-worker/src/transaction_publisher/publisher.service.ts
+++ b/apps/account-worker/src/transaction_publisher/publisher.service.ts
@@ -7,7 +7,7 @@ import { SubmittableExtrinsic } from '@polkadot/api-base/types';
 import { IMethod, ISubmittableResult, Signer, SignerResult } from '@polkadot/types/types';
 import { SchedulerRegistry } from '@nestjs/schedule';
 import { BlockchainService } from '#blockchain/blockchain.service';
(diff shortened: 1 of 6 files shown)


mattheworris (approved):
- Read through changes
🚢 it!

wilwade (approved):
🎉

PR #771, opened 2025-03-31
fix: increase default incoming request api timeout to 30s

JoeCap08055:
Increase the default timeout for incoming http requests to 30 seconds, and update documentation accordingly.

[Note, due to an errant typo that changed the default API port for `content-watcher` instead of the API timeout, I went down a rabbit-hole of debugging the K6 load test pipeline. In the process I hit upon a CI optimization that should speed up the pipeline a bit by building only a single image for both api & worker services, included in this PR]


wilwade (approved):
❤️

PR #776, opened 2025-04-01
Fix/account load test issues

### LINKED ISSUE #774: [bug] [account-service] Various issues from load testing
# Description
Load testing the SIWF sign-up flow in `account-service` has revealed the following:

- [x] Setting the transaction publish queue concurrency > 1 results in transaction errors
- [x] Block header logic for signing block is reversed
- [x] Some transaction failures should be retried (with a limit)
    - `BadProof`

JoeCap08055:
- Revert the transaction publish queue default concurrency to 1, until we do some more investigation
- Fix inverted signing block header logic
- Retry failed chain transactions up to 4 times for the following RPC errors:
    - bad signature (Invalid Transaction)
    - insufficient balance
    - nonce conflict
    - expired mortality
- Added some happy path logging to account-service for the SIWFv2 signup flow to aid in troubleshooting

Closes #774

### DIFF
--- a/apps/account-worker/src/transaction_publisher/publisher.service.ts
+++ b/apps/account-worker/src/transaction_publisher/publisher.service.ts
@@ -131,7 +131,8 @@ export class TransactionPublisherService extends BaseConsumer implements OnAppli
           throw new Error(`Invalid job type.`);
         }
       }
-      this.logger.debug(`Successful job: ${JSON.stringify(job, null, 2)}`);
+      this.logger.debug(`Job successfully completed (${job.id})`);
+      this.logger.verbose(JSON.stringify(job, null, 2));
 
       const status: ITxStatus = {
         type: job.data.type,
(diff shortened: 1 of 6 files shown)


libs/blockchain/src/blockchain.service.ts
```diff
@@ -198,6 +204,30 @@ export class BlockchainService extends BlockchainRpcQueryService implements OnAp
     }
   }
 
+  public static createError(err: any): any {
```
wilwade, line 207: Nice function!

mattheworris (approved):
🚢 it!

apps/account-worker/src/transaction_publisher/publisher.service.ts
```diff
@@ -149,6 +155,29 @@ export class TransactionPublisherService extends BaseConsumer implements OnAppli
...
+   * so we cap it at 4)
+   * @param job The current job
+   * @param err The error that was thrown
+   * @returns Whether the job should be retried
+   */
+  public static shouldRetry(job: Job<any>, err: Error): boolean {
+    if (
+      job.attemptsStarted < 4 &&
+      (err instanceof NonceConflictError ||
+        err instanceof BadSignatureError ||
```
aramikm, line 171: If we get the signature from the user then retrying a bad signature would not work but that might not be the case in here

wilwade: I believe that is a different error

JoeCap08055: The bad signature is the provider submitting the transaction, not the payload inside the transaction... that would, I think, be an extrinsic error on-chain.

wilwade (approved):
Adding approval for the logs

PR #785, opened 2025-04-02
fix: add configurable Redis options and default commandTimeout

JoeCap08055:
This PR adds some configurability to our Redis clients and defaults some options.
* Add `REDIS_OPTIONS` env var (optional). Must be a JSON object string, mapping to a `RedisOptions` object [defined here](https://github.com/redis/ioredis/blob/2ed6414339bdda3e427ed1e8484263cc4848f2ac/lib/redis/RedisOptions.ts)
* Default Redis `commandTimeout` to 5s if not overriden by `REDIS_OPTIONS`
* Update all app modules to use new Redis options config
* Update docs, env templates


env-files/account.template.env
```diff
@@ -33,6 +33,10 @@ WEBHOOK_BASE_URL=http://host.docker.internal:3001/webhooks/account-service
 REDIS_URL=redis://0.0.0.0:6379
 #REDIS_URL=redis://redis:6379                        #docker dev
 
+# Additional Redis client options as a JSON object string
+# See https://github.com/redis/ioredis/blob/2ed6414339bdda3e427ed1e8484263cc4848f2ac/lib/redis/RedisOptions.ts
+REDIS_OPTIONS='{"commandTimeout":5000}'
```
wilwade, line 38: I think we want to put the commandTimeout in the code defaults. That way if someone just wants to set some other option, they don't lose the default commandTimeout

wilwade: Ah I see. This is example and it is also set in `DEFAULT_REDIS_OPTIONS`. I missed that. No trouble here then.

wilwade (approved):
:shipit:

mattheworris (approved):
🚢 it!

PR #792, opened 2025-04-03
Consolidate Redis options and handle ssl

JoeCap08055:
# Description
* Consolidate all Redis options, whether URL or JSON, to a normalized `RedisOptions` config object
* Make config `REDIS_URL` and `REDIS_OPTIONS` mutually exclusive
* Eliminate now-redundant `QueueConfig`, as the queue module can now just use `CacheConfig` directly
* Handle `rediss://` protocol in `REDIS_URL` to propagate correctly to `RedisOptions` object
* Update docs

## Additional fixes
* Add `redis/redis.conf` default config file to be used by local Docker 'redis' service when testing
* Remove the additional key address from individual nonce keys in Redis (added in #758); the address is already baked into the cache prefix for that connection
* When running the 'start.sh' script, have it ask for & optionally start the mock webhook logger
* When running the 'stop.sh' script, instead of trying to delete specific Docker volumes individually, have it just use the `--volumes` option to `docker compose down`; will auto-remove any volumes created
* Normalize config validation; previously errors were reported with the Javascript property names instead of environment variable names, which was confusing for the user


developer-docs/content-watcher/ENVIRONMENT.md
```diff
@@ -3,24 +3,24 @@
...
+| `IPFS_ENDPOINT`                    | URL to IPFS endpoint                                                                                            |                            URL                            |                  Y                   |                            |
+| `IPFS_GATEWAY_URL`                 | IPFS gateway URL '[CID]' is a token that will be replaced with an actual content ID                             |                       URL template                        |                  Y                   |                            |
+| `QUEUE_HIGH_WATER`                 | Max number of jobs allowed on the queue before blockchain scan will be paused to allow queue to drain           |                          >= 100                           |                                      |            1000            |
+| `REDIS_OPTIONS`                    | Additional Redis options.<br/>See https://ioredis.readthedocs.io/en/latest/API/#new-redisport-host-options      |                        JSON string                        |   Y<br/>(either this or REDIS_URL)   | '{"commandTimeout":10000}' |
```
wilwade, line 21: Suggestion for the link
```suggestion
| `REDIS_OPTIONS`                    | Additional Redis options.<br/>See [ioredis configuration](https://ioredis.readthedocs.io/en/latest/API/#new-redisport-host-options)      |                        JSON string                        |   Y<br/>(either this or REDIS_URL)   | '{"commandTimeout":10000}' |
```

env-files/account.template.env
```diff
@@ -30,12 +30,14 @@ ACCOUNT_WEBHOOK_PORT=3001
...
 REDIS_URL=redis://0.0.0.0:6379
 #REDIS_URL=redis://redis:6379                        #docker dev
 
-# Additional Redis client options as a JSON object string
-# See https://github.com/redis/ioredis/blob/2ed6414339bdda3e427ed1e8484263cc4848f2ac/lib/redis/RedisOptions.ts
-REDIS_OPTIONS='{"commandTimeout":5000}'
+# Redis client options as a JSON object string.
+# Only one of REDIS_URL or REDIS_OPTIONS may be specified.
+# See https://ioredis.readthedocs.io/en/latest/API/#new-redisport-host-options
+#REDIS_OPTIONS='{"commandTimeout":10000}'
```
wilwade, line 40: We should give in the example setting the redis host since that would be required since you cannot use the REDIS_URL.

JoeCap08055: Updated.

apps/account-api/src/api.config.ts
JoeCap08055, line 1: Changed all `*.config.ts` files to use config name normalization so that errors will be reported with the env variable name instead of the JS property name; less confusing for users.

libs/blockchain/src/blockchain.service.ts
```diff
@@ -55,11 +55,11 @@ export interface IHeaderInfo {
 
 const { NUMBER_OF_NONCE_KEYS_TO_CHECK, NONCE_KEY_EXPIRE_SECONDS, getNonceKey } = NonceConstants;
 
-function getNextPossibleNonceKeys(account: string, currentNonce: string | number): string[] {
+function getNextPossibleNonceKeys(currentNonce: string | number): string[] {
   const keys: string[] = [];
   for (let i = 0; i < NUMBER_OF_NONCE_KEYS_TO_CHECK; i += 1) {
     const key = Number(currentNonce) + i;
-    keys.push(getNonceKey(account, `${key}`));
+    keys.push(getNonceKey(`${key}`));
```
JoeCap08055, line 62: Reverted this change from a previous PR; turns out we already include the address in the cache key prefix.

start.sh
```diff
@@ -158,6 +158,9 @@ EOI
     if yesno "Start the account service" Y; then
         PROFILES="${PROFILES} account"
     fi
+    if yesno "Start the mock webhook logger" Y; then
+        PROFILES="${PROFILES} webhook"
+    fi
```
JoeCap08055, line 163: Added this to eliminate delays and error messages in `account-service-worker` due to a webhook URL being specified (mandatory), but nothing listening at the specified address.

stop.sh
```diff
@@ -50,29 +50,16 @@ COMPOSE_CMD=$( prefix_postfix_values "${COMPOSE_FILES}" "-f ")
...
-    docker volume rm -f ${COMPOSE_PROJECT_NAME}_graph_worker_node_cache
-    docker volume rm -f ${COMPOSE_PROJECT_NAME}_content_publishing_api_node_cache
-    docker volume rm -f ${COMPOSE_PROJECT_NAME}_content_publishing_worker_node_cache
-    docker volume rm -f ${COMPOSE_PROJECT_NAME}_content_watcher_node_cache
-    docker volume rm -f ${COMPOSE_PROJECT_NAME}_gateway_base_node_cache
-    if [[ ! $TESTNET_ENV == "true" ]]
-    then
-        docker volume rm -f ${COMPOSE_PROJECT_NAME}_chainstorage
-    fi
+    VOLUMES="--volumes"
```
JoeCap08055, line 58: Turns out adding `--volumes` to a `docker compose down` command will remove the relevant volumes. Also eliminates the issue of unnamed volumes not being cleaned up.

libs/cache/src/cache.config.spec.ts
```diff
@@ -1,7 +1,10 @@
+/* eslint-disable no-underscore-dangle */
 /* eslint-disable import/no-extraneous-dependencies */
 import { describe, it, expect, beforeAll } from '@jest/globals';
 import cacheConfig, { ICacheConfig } from './cache.config';
 import configSetup from '#testlib/utils.config-tests';
+import { url } from 'inspector';
+import { urlencoded } from 'express';
```
mattheworris, line 7: VSCode says these imports are unused and the tests work without them.
```suggestion
```

libs/cache/src/cache.config.ts
```diff
@@ -1,35 +1,68 @@
 import { JoiUtils } from '#config';
 import { registerAs } from '@nestjs/config';
 import { RedisOptions } from 'ioredis';
+import { parseURL } from 'ioredis/built/utils';
 import Joi from 'joi';
 
 export interface ICacheConfig {
-  redisUrl: string;
+  // redisUrl: string;
```
mattheworris, line 8: nit: remove commented out code
```suggestion
```

libs/cache/src/cache.config.spec.ts
mattheworris, line 7: VSCode says these imports are unused and the tests work without them.
```suggestion
```

mattheworris (approved):
Nice work.
🚢 it!

PR #811, opened 2025-04-18
get all graphs for a certain MsaId & Add logs for all axios call

prestonzendev:
# Solution

- Get all graphs for a Certain MsaId
- Add logs for all axios Calls
- Set httpResponseTimeoutMS based on API_TIMEOUT_MS


apps/account-api/test/setup/test-revoke-e2e.ts
```diff
@@ -26,7 +26,34 @@ async function revokeDelegation() {
   const { keypair } = users[1];
   const accountId = keypair.address;
   const getPath: string = `http:/localhost:3013/v1/delegation/revokeDelegation/${accountId}/${providerId}`;
-  const response = await axios.get(getPath);
+  console.log(`[Test] Making GET request to: ${getPath}`);
```
wilwade, line 29: This is an e2e test. Not sure this is supposed to be here.

apps/account-worker/src/transaction_notifier/notifier.service.ts
```diff
@@ -233,15 +233,42 @@ export class TxnNotifierService
...
               await axios.post(webhook, webhookResponse, { timeout: this.httpConfig.httpResponseTimeoutMS });
               this.logger.debug(`Transaction Notification sent to webhook: ${webhook}`);
               break;
             } catch (error) {
-              this.logger.error(`Failed to send notification to webhook: ${webhook}`);
-              this.logger.error(error);
+              this.logger.error(`[Attempt ${retries + 1}] Failed to send notification to webhook: ${webhook}`);
+              
+              if (axios.isAxiosError(error)) {
+                this.logger.error('Axios Error Details:');
```
wilwade, line 247: This sort of multi-line error message breaks a lot of log collection systems.

libs/account-lib/src/services/provider-webhook.service.ts
```diff
@@ -40,6 +40,60 @@ export class ProviderWebhookService implements OnModuleDestroy {
...
+          status: response.status,
+          statusText: response.statusText,
+          headers: response.headers,
+          data: response.data
+        });
+        return response;
+      },
+      (error) => {
+        if (axios.isAxiosError(error)) {
+          this.logger.error('[Provider] Response interceptor error:');
```
wilwade, line 76: Seeing all this repeated, it feels like we should have a logger helper function.

libs/config/src/http-common.config.ts
```diff
@@ -9,7 +9,8 @@ export interface IHttpCommonConfig {
 export default registerAs('http-common', (): IHttpCommonConfig => {
   const configs: JoiUtils.JoiConfig<IHttpCommonConfig> = JoiUtils.normalizeConfigNames({
     httpResponseTimeoutMS: {
-      label: 'HTTP_RESPONSE_TIMEOUT_MS',
+      value: process.env.HTTP_RESPONSE_TIMEOUT_MS &&
+        Math.max(Number(process.env.HTTP_RESPONSE_TIMEOUT_MS), Number(process.env.API_TIMEOUT_MS ?? 0)),
```
wilwade, line 13: This `Math.max` would break in the case where `API_TIMEOUT_MS` is set lower than `HTTP_RESPONSE_TIMEOUT_MS`, but `API_TIMEOUT_MS` should always be less than `HTTP_RESPONSE_TIMEOUT_MS`

wilwade (changes requested):
1. We need to split this up into two PRs.
2. I didn't review the Graph part of it
3. A good amount of `npm run format` needed

Issue #800, opened 2025-04-09
Graph service (enhancements)


p5150j:

prestonzen:
- Any breaking changes make a V2 Graph Service controller instead of updating V1
- Clarify naming convention in the service
- Reference rust graph SDK dependency of how the graph is stored on the Frequency chain for easy API route creation: https://github.com/ProjectLibertyLabs/graph-sdk

PR #813, opened 2025-04-18
get all graphs for a certain msaid

prestonzendev:
# Solution

create getAllGraphsForMsaId to get all graphs for a certain MsaId


wilwade (changes requested):
I'm not sure this was the goal of the issue per #800

So:
1. What part of #800 was #804 supposed to accomplish?
2. At a minimum this code isn't used anywhere
3. Needs tests
4. Not sure how this is different from `getGraphs`

Issue #756, opened 2025-03-24
[bug] [content-publishing] DELETE request should not have a body


JoeCap08055:
# Description
The route `DELETE /v1/content/{msaId}` to submit a Tombstone request currently expects a body with the rest of the parameters of the Tombstone request. This is problematic because the [HTTP specification](https://www.rfc-editor.org/rfc/rfc9110.html#section-9.3.5) indicates that there are no semantics defined for the body of a DELETE request, and therefore implementations may explicitly reject such content.

Also, technically, since a Tombstone request, while it is the mechanism one would use to indicate a wish to have  previously-posted content "deleted", it is not actually or semantically a "delete" of an allocated web resource; to the contrary, it actually _creates_ a new resource (a Tombstone notification).

# Acceptance Criteria
- [ ] Create a new, `POST /v2/content/{msaId}/tombstones` endpoint to handle this use case
- [ ] mark or schedule the existing `DELETE /v1/content/{msaId}` endpoint for deprecation 
- [ ] API docs & generated types re-generated as necessary

emresurmeli:
@JoeCap08055  Taking a look at this and I think I might be able to pick it up.

Couple questions:

1. Does it make sense to follow the same pattern as the deprecation the `/v1/asset/upload` PUT route underwent? 

2. We might also want to add updating the [docs](https://projectlibertylabs.github.io/gateway/Build/ContentPublishing/Api.html) to the Acceptance Criteria? 

I might be missing some other things I'm not aware of.

emresurmeli:
Opened a WIP PR #816

PR #816, opened 2025-04-25
Content publishing; DELETE /v1/content/:msaId deprecation and new POST /v2/content/:msaId/tombstones

emresurmeli:
Solving #756 

I've followed the deprecation pattern that was used in `/v1/asset/upload` however I was unsuccessful when testing the routes. Seems I wasn't able to setup my gateway properly locally, will continue to work on doing that but I'll drop the code here if someone wants to test it out on their end.

Still left to do:
1. Write tests
2. Generate API docs 

cc: @JoeCap08055

### DIFF
--- a/apps/content-publishing-api/src/controllers/v2/content.controller.v2.ts
+++ b/apps/content-publishing-api/src/controllers/v2/content.controller.v2.ts
@@ -77,7 +77,10 @@ export class ContentControllerV2 {
   @ApiOperation({ summary: 'Delete DSNP Content for user' })
   @HttpCode(202)
   @ApiResponse({ status: '2XX', type: AnnouncementResponseDto })
-  async delete(@Param() { msaId }: MsaIdDto, @Body() tombstoneDto: TombstoneDto): Promise<AnnouncementResponseDto> {
+  async postTombstone(
+    @Param() { msaId }: MsaIdDto,
+    @Body() tombstoneDto: TombstoneDto,
+  ): Promise<AnnouncementResponseDto> {
     return this.apiService.enqueueRequest(AnnouncementTypeName.TOMBSTONE, msaId, tombstoneDto);
   }
 }
--- a/apps/content-publishing-api/src/controllers/v2/content.controller.v2.ts
+++ b/apps/content-publishing-api/src/controllers/v2/content.controller.v2.ts
@@ -12,7 +12,8 @@ import {
 } from '@nestjs/common';
 import { ApiOperation, ApiResponse, ApiTags } from '@nestjs/swagger';
 import { ApiService } from '../../api.service';
-import { AnnouncementResponseDto, BatchFilesDto, OnChainContentDto } from '#types/dtos/content-publishing';
+import { AnnouncementResponseDto, BatchFilesDto, OnChainContentDto, TombstoneDto } from '#types/dtos/content-publishing';
+import { AnnouncementTypeName } from '#types/enums';
 import { BlockchainRpcQueryService } from '#blockchain/blockchain-rpc-query.service';
 import { MsaIdDto } from '#types/dtos/common';
 import apiConfig, { IContentPublishingApiConfig } from '#content-publishing-api/api.config';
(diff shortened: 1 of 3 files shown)


.github/workflows/build.yml
```diff
@@ -163,7 +163,7 @@ jobs:
           buildkitd-config-inline: |
             [registry."ghcr.io"]
       - name: Login to DockerHub
-        if: ${{ steps.should.outputs.RUN == 'true' }}
+        if: ${{ steps.should.outputs.RUN == 'true' && secrets.DOCKERHUB_USERNAME != '' && secrets.DOCKERHUB_TOKEN != '' }}
```
wilwade, line 166: I added this, as it is just for rate limited reasons, but forks don't need to have access to the docker hub creds.

apps/content-publishing-api/src/controllers/v2/content.controller.v2.ts
```diff
@@ -71,4 +77,15 @@ export class ContentControllerV2 {
     const promises = batchContentDto.batchFiles.map((batchFile) => this.apiService.enqueueBatchRequest(batchFile));
     return Promise.all(promises);
   }
+
+  @Post(':msaId/tombstones')
+  @ApiOperation({ summary: 'Delete DSNP Content for user' })
+  @HttpCode(202)
+  @ApiResponse({ status: '2XX', type: AnnouncementResponseDto })
+  async postTombstone(
```
wilwade, line 85: @emresurmeli Could you add a simple test for this in `apps/content-publishing-api/test/app.e2e-spec.ts` (really need to split that up, but another time).

Should be able to mostly just copy/paste the one for the older endpoint

.github/workflows/build.yml
```diff
@@ -162,8 +162,19 @@ jobs:
...
         if: ${{ steps.should.outputs.RUN == 'true' }}
+        run: |
+          if [ -n "${{ secrets.DOCKERHUB_USERNAME }}" ] && [ -n "${{ secrets.DOCKERHUB_TOKEN }}" ]; then
+            echo "do_dockerhub_login=true" >> $GITHUB_OUTPUT
+          else
+            echo "do_dockerhub_login=false" >> $GITHUB_OUTPUT
+          fi
+
+      - name: Login to DockerHub
+        if: ${{ steps.should.outputs.RUN == 'true' && steps.docker-credentials.outputs.do_dockerhub_login == 'true' }}
```
wilwade, line 177: I added this, as it is just for rate limited reasons, but forks don't need to have access to the docker hub creds.

apps/content-publishing-api/src/controllers/v2/content.controller.v2.ts
emresurmeli: Added, emulated the deprecated DELETE endpoint

wilwade (commented):
Looks good from my side. I'll have @JoeCap08055 give it a final once over and we'll get it merged! Thanks for the contribution!

.github/workflows/e2e-tests.yml
```diff
@@ -16,7 +16,7 @@ jobs:
       # NOTE: only account E2E tests enabled in pipeline currently; add other services as they are updated
       # IF you add a new filter it ALSO needs to be here
       services: >-
-        ["account"]
+        ["account", "content-publishing"]
```
JoeCap08055, line 19: I added this, now that these tests are working

JoeCap08055 (changes requested):
Looks good--I fixed some long-standing issues preventing the E2E tests for this app not to work.
Please add a section in the K6 load test script for this new endpoint.

apps/content-publishing-api/src/controllers/v2/content.controller.v2.ts
```diff
@@ -71,4 +77,15 @@ export class ContentControllerV2 {
     const promises = batchContentDto.batchFiles.map((batchFile) => this.apiService.enqueueBatchRequest(batchFile));
     return Promise.all(promises);
   }
+
+  @Post(':msaId/tombstones')
+  @ApiOperation({ summary: 'Delete DSNP Content for user' })
```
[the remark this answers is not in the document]
JoeCap08055: I committed an update to this description.

emresurmeli:
@JoeCap08055 I added the k6 test, which I emulated based on the other tests by creating a valid tombstone body in the `mockRequestData.ts` file. Let me know if what you had in mind was something else!

PR #821, opened 2025-04-29
Attempt to improve speed of IPFS publishing by checking for CID existence via gatewayUrl

wesbiggs:
# Problem

With a large pinset, `ipfs.pin.ls` can be unacceptably slow, and might block other IPFS requests.

# Solution

Changes the behavior of IpfsService methods `getPinned` and `getInfo` to make a HTTP HEAD request via the configured `ipfsGatewayUrl` instead of calling `ipfs.pin.ls`.

A new method, `checkExists`, was added. It uses the [`Cache-Control: Only-If-Cached`](https://specs.ipfs.tech/http-gateways/path-gateway/#only-if-cached) header to ensure the file is not looked for on the wider IPFS network.

The old method, `isPinned`, was not removed, but is currently unused.

## Steps to Verify:

1. Ensure content publishing service is working correctly.
2. Analyze performance.


wilwade (commented):
I'll leave the final approval to @JoeCap08055, but this looks good to me. @wesbiggs I'm ok with removing the ls function as well since that function open to potential misuse.

JoeCap08055 (approved):
Looks good.
Unfortunately, I don't think this will solve our pinning performance problems, as the `pin.ls` call, while a tad slow when the pinned set is large, is not the main source of our bottleneck; added a new pinned file is the main problem.

That said, every little optimization helps!

libs/storage/src/ipfs/ipfs.service.ts
```diff
@@ -70,6 +73,17 @@ export class IpfsService {
     return response;
   }
 
+  public async checkExists(cid: string): Promise<boolean> {
```
aramikm, line 76: is this safe to use from durability perspective since we might have files that do exist but are not pinned!?

JoeCap08055: Hmmm... for the `content-publishing` use case, at least, we expect that any file we're querying was recently uploaded by the same content-publishing instance to the same IPFS node, so if the local node doesn't have it, it probably doesn't exist.

However, you bring up a good point for naming, at least--perhaps:
```suggestion
  public async checkExistsInLocalGateway(cid: string): Promise<boolean> {
```

aramikm: I might be getting this completely wrong but my understanding is that there exists a state that you uploaded a file but that upload would be only temporary unless you pin it. So if this assumption is correct then is there a possibility that you would find that file locally but it's not pinned in reality and it can get removed by node at any minute.

JoeCap08055: * For content-publishing, we don't just "upload", we pin. Though, we don't "care" if it's pinned (though it makes sense for it to be so; we only need to verify that it was uploaded & available currently). Whether any content remains pinned after that is not guaranteed; ie, as a provider I could remove pins whenever I want to manage my IPFS resources.
* For content-watcher, we don't care if it's pinned at all, just that it's available at the time we discover it.

So, for content-publishing purposes, we upload & pin (because we want the content to remain available), but from a workflow perspective we are mainly interested that the upload was successful & the resource is currently available. Because even if it's pinned right now, that pin could be removed at any point that the owner of the node so chooses.

PR #826, opened 2025-05-06
fix: add 'v2' endpoints to generated docs

### LINKED ISSUE #825: [content-publishing] Generated docs missing 'v2' endpoints
# Description
The generated Swagger docs for `content-publishing-api` are missing the `v2` endpoints

JoeCap08055:
# Description
Added the `v2` endpoints to the generated Swagger docs for `content-publishing-api`

Closes #825


wilwade (approved):
📎

mattheworris (approved):

JoeCap08055:
Turns out the recent dependabot update from `@nestjs/common@10.4.15` to `10.4.17` contained a breaking change that was causing both our E2E and K6 tests to fail.
Nature of the change: NestJS now depends on `file-type` and the ParseFilePipe now examines the file bytes for the "magic number" header by default and does not merely trust the MIME type information (if any) sent by the client, or the file extension.

## Problem # 1
`file-type` is an ESM module, and Jest is unable to load it in this CommonJS project, so the E2E test were failing.
* Solution: detect the Jest environment, and enable the skipMagicNumbersValidation option

## Problem # 2
K6 tests against the live app, but the file content it uses to test with is garbage.
* Solution: bake the appropriate initial bytes into the test content. For now, we support JPG, OGG, and Parquet content. More can be added if tests require it.

## Problem # 3
The `file-type` package translates Parquet files as the MIME type 'application/x-parquet`, instead of the official `application/vnd.apache.parquet`.
* Solution: accept both

PR #828, opened 2025-05-07
Announce correct batch file size

wesbiggs:
# Problem

`ipfs.dag.stat` gives a `size` value that is not the same as the actual size of a file when downloaded.
This is because the DAG stores additional info beyond the raw file bytes.

An example can be seen in the difference between `dag.stat` and `files.stat`, where the actual length of the file is 87,699 bytes:

`dag.stat` reports the incorrect (for our purposes) value (twice, for good measure):

```
% curl -s -d '' '$IPFS_RPC_URL/api/v0/dag/stat?arg=bafybeibkvwtaugltazjt7oeq6xahz2qwvd5sm75twsht3e45niuxnfg654' | jq .
{
  "UniqueBlocks": 1,
  "TotalSize": 87713,
  "Ratio": 1,
  "DagStats": [
    {
      "Cid": "bafybeibkvwtaugltazjt7oeq6xahz2qwvd5sm75twsht3e45niuxnfg654",
      "Size": 87713,
      "NumBlocks": 1
    }
  ]
}
```

... whereas `files.stat` gives us both the file size and the DAG size:

```
% curl -s -d '' '$IPFS_RPC_URL/api/v0/files/stat?arg=/ipfs/bafybeibkvwtaugltazjt7oeq6xahz2qwvd5sm75twsht3e45niuxnfg654&progress=true' | jq .
{
  "Hash": "bafybeibkvwtaugltazjt7oeq6xahz2qwvd5sm75twsht3e45niuxnfg654",
  "Size": 87699,
  "CumulativeSize": 87713,
  "Blocks": 0,
  "Type": "file"
}
```

The correct value is also present in a `HEAD` request via the gateway:

```
% curl -s --head '$IPFS_GATEWAY_URL/ipfs/bafybeibkvwtaugltazjt7oeq6xahz2qwvd5sm75twsht3e45niuxnfg654' | grep Content-Length:
Content-Length: 87699
```

# Solution

Since we can get the correct size from the `Content-Length` header of the `HEAD /ipfs/[CID]` response, and we're already using this as an existence check, I modified the code to get size info from this request as well.

It has the nice side effect of one less request to the IPFS server.

The `ipfsService.getInfo(cid)` method has been removed since it was otherwise unused, and was not providing the output expected. The batch announcer has been updated to get the file size value directly via the new method, `ipfsService.contentLengthInLocalGateway(cid)`.

## Steps to Verify:

1. Check the `payload_length` reported in a Frequency IpfsMessage against the actual file size of the batch.

### DIFF
--- a/libs/storage/src/ipfs/ipfs.service.ts
+++ b/libs/storage/src/ipfs/ipfs.service.ts
@@ -79,9 +79,12 @@ export class IpfsService {
   }
 
   public async existsInLocalGateway(cid: string): Promise<boolean> {
-    return this.contentLengthInLocalGateway(cid)
-      .then(() => true)
-      .catch(() => false);
+    try {
+      await this.contentLengthInLocalGateway(cid);
+      return true;
+    } catch (_err) {
+      return false;
+    }
   }
 
   public async isPinned(cid: string): Promise<boolean> {
--- a/libs/storage/src/ipfs/ipfs.service.ts
+++ b/libs/storage/src/ipfs/ipfs.service.ts
@@ -6,7 +6,7 @@ import { randomUUID } from 'crypto';
 import { extension as getExtension } from 'mime-types';
 import { FilePin } from '#storage/ipfs/pin.interface';
 import { calculateDsnpMultiHash, calculateIncrementalDsnpMultiHash } from '#utils/common/common.utils';
-import { createKuboRPCClient, KuboRPCClient, CID, BlockStatResult } from 'kubo-rpc-client';
+import { createKuboRPCClient, KuboRPCClient, CID } from 'kubo-rpc-client';
 import httpCommonConfig, { IHttpCommonConfig } from '#config/http-common.config';
 import { Readable } from 'stream';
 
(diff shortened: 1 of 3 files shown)


aramikm (approved):
Good find! Approved!

libs/storage/src/ipfs/ipfs.service.ts
```diff
@@ -59,29 +59,27 @@ export class IpfsService {
...
       method: "HEAD",
       headers: {
         "Cache-Control": "only-if-cached",
       },
     });
-    return response && response.status === 200;
+    if (response && response.status === 200) {
+      return Number(response.headers.get("Content-Length"));
+    } else {
+      return -1;
```
JoeCap08055, line 77: Hmm, since the old function was throwing an error if the resource doesn't exist, perhaps we want to preserve that behavior?

wesbiggs: Good call, will update.

libs/storage/src/ipfs/ipfs.service.ts
```diff
@@ -60,26 +60,28 @@ export class IpfsService {
...
+      return Number(response.headers?.get('Content-Length'));
     }
+    throw new Error(`Unable to access HEAD for ${cid}`);
   }
 
   public async existsInLocalGateway(cid: string): Promise<boolean> {
-    return -1 !== (await this.contentLengthInLocalGateway(cid));
+    return this.contentLengthInLocalGateway(cid)
+      .then(() => true)
+      .catch(() => false);
```
JoeCap08055, line 84: nit: I find, stylistically, that mixing idioms gets confusing to readers/maintainers (ie, either use "Promise.then().catch()" or "await + try/catch". Gateway uses the async/await model throughout, so let's stick to that here.
```suggestion
try {
    await this.contentLengthInLocalGateway(cid);
    return true;
} catch (_err) {
    return false;
}
```

PR #829, opened 2025-05-14
logging header data, filestream and req events

shannonwells:
# Problem
Debugging logs for some freesky issues

# Solution
Adding some logging for:
* request Content-length header
* request events end, close, aborted
* Inside of busboy handler:
    * whether the filestream 'truncated' flag is set
    * filestream "end" event inside of the busboy handler

Also:
* small correction to README, 
* v2/assets/upload e2e test to verify above,
* rename load test b/c it was missed and not being run in CI

with @JoeCap08055 

To verify, run content-publisher with DEBUG=true in environment file and publish an asset. The debugging prints should be logged.


apps/content-publishing-api/src/controllers/v2/asset.controller.v2.ts
```diff
@@ -32,9 +32,13 @@ export class AssetControllerV2 {
   })
   @ApiResponse({ status: '2XX', type: UploadResponseDtoV2 })
   async uploadFile(@Req() req: Request, @Res({ passthrough: true }) _res: Response): Promise<IUploadResponse> {
-    const busboy = Busboy({ headers: req.headers });
+    req.on('end', () => console.log('Request ended'));
+    req.on('close', () => console.log('Request closed'));
+    req.on('aborted', () => console.log('Request aborted'));
```
JoeCap08055, line 37: If these are to be permanent, probably want `this.logger.verbose`

apps/content-publishing-api/test/app-v2-asset-upload-e2e.spec.ts
JoeCap08055, line 1: nit: lots of duplication here; could the new test simply be added to the existing `app.e2e-spec.ts` file? If you think that file is getting too big/unwieldy, I think a future issue to refactor into common setup & domain-specific endpoint test files would be in order.

JoeCap08055: Also, if we're adding E2E tests, probably should add one for `/v2/content/batchAnnouncement`, seeing as that's what's being heavily used on testnet right now.

shannonwells: I think we should leave other tests for another PR since the idea here was just to verify the logging stuff works, and get that pushed out.

JoeCap08055 (approved):

PR #831, opened 2025-05-16
fix: better file stream error handling

JoeCap08055:
# Problem
`content-publishing-api` `POST /v2/asset/upload` endpoint has been experiencing unhandled file stream termination errors. Although we have been unable to reproduce locally to determine the root cause, code inspection reveals that we are not handling errors properly on the file streams.

This solution may not eliminate the errors & still result in assets that fail to upload; however, it should at least prevent the application from crashing, and possibly allow for some better data collection as a result.

### DIFF
--- a/apps/content-publishing-api/src/api.service.ts
+++ b/apps/content-publishing-api/src/api.service.ts
@@ -309,6 +309,10 @@ export class ApiService {
     uploadPassThru.on('error', handleError);
     hashPassThru.on('error', handleError);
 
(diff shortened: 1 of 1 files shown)


wesbiggs:
Do you have access to any byte counts that could be logged? That could help pin down if it always happens at a certain size threshold (or time threshold, maybe).

mattheworris (approved):
- Read through changes
🚢 it!

PR #833, opened 2025-05-20
fix: get correct file size from IPFS, without hanging

### LINKED ISSUE #832: [bug] IPFS file size not being reported correctly
# Description
We currently use a `HTTP HEAD` method to get the file size of a file in IPFS (if it exists). However, depending on HTTP gateway configurations and other factors, this can sometimes report a file size of zero, even for a file that exists in the local IPFS node. However, other methods of retrieving file size must be used with care, because if the file is not cached/pinned by the local node, the request could hang (or at least, appear to be hung due to reaching out to the global IPFS network to attempt to get the file information).

# Acceptance Criteria
- File size is retrieved/reported correctly
- Request to retrieve file size/info does not hang.

JoeCap08055:
# Description
This PR restores the logic to get the file size of a file in IPFS using the Kubo RPC API, with the added logic to avoid hanging if the file is not cached in the local node.

Closes #832

### DIFF
--- a/apps/content-publishing-worker/src/batch_announcer/batch.announcer.ts
+++ b/apps/content-publishing-worker/src/batch_announcer/batch.announcer.ts
@@ -73,13 +73,13 @@ export class BatchAnnouncer {
     // Get previously uploaded file from IPFS
     this.logger.log(`Getting info from IPFS for ${batch.cid}`);
     try {
-      const size = await this.ipfsService.contentLengthInLocalGateway(batch.cid);
-      this.logger.debug(`Got info from IPFS: size=${size}`);
+      const { cid, size } = await this.ipfsService.getInfoFromLocalNode(batch.cid);
+      this.logger.debug(`Got info from IPFS: cid=${cid}, size=${size}`);
 
       const response = {
         id: batch.cid,
         schemaId: batch.schemaId,
-        data: { cid: batch.cid, payloadLength: size },
+        data: { cid: cid.toV1().toString(), payloadLength: size },
       };
       this.logger.debug(`Created job to announce existing batch: ${JSON.stringify(response)}`);
       return response;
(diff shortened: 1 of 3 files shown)


libs/storage/src/ipfs/ipfs.service.ts
```diff
@@ -59,32 +59,34 @@ export class IpfsService {
     return Buffer.concat(chunks);
   }
 
-  /**
-   * gets the true content-length of the CID file from the IPFS gateway,
-   * or NaN if the content-length is unknown.
-   * if the file does not exist locally, or a network error occurs, throws an error
-   */
-  public async contentLengthInLocalGateway(cid: string): Promise<number> {
+  public async getInfoFromLocalNode(cid: string): Promise<FilesStatResult> {
```
JoeCap08055, line 62: (Took out the `checkForExistence` flag here, because this will now error (not hang) if the file doesn't exist in the local node, so we can save the extra network call)

As noted in the comment, the Kubo CLI maps 1:1 to the Kubo RPC API, with all supported CLI options mapped to properties... but for some reason, the JS package doesn't include a specific property for the `offline` option. We can explicitly override typing here as we do, or we could use the `timeout: number | string` property (milliseconds) to achieve the desired result.

apps/content-publishing-worker/src/batch_announcer/batch.announcer.ts
```diff
@@ -73,13 +73,13 @@ export class BatchAnnouncer {
...
-      const size = await this.ipfsService.contentLengthInLocalGateway(batch.cid);
-      this.logger.debug(`Got info from IPFS: size=${size}`);
+      const { cid, size } = await this.ipfsService.getInfoFromLocalNode(batch.cid);
+      this.logger.debug(`Got info from IPFS: cid=${cid}, size=${size}`);
 
       const response = {
         id: batch.cid,
         schemaId: batch.schemaId,
-        data: { cid: batch.cid, payloadLength: size },
+        data: { cid: cid.toV1().toString(), payloadLength: size },
```
shannonwells, line 82: Informational only: Are we only ever going to support v1 CIDs?  we were discussing CIDv2 at some point.

JoeCap08055: The issue is, CIDv1 is deterministic based solely on the hashed file contents. CIDv2 is additionally dependent on the chunking parameters with which the file was uploaded/stored on IPFS. So: we can pre-compute a CIDv1 based on the file contents, but any CIDv2 we attempt to pre-compute will not necessarily match the CIDv2 that IPFS generates for content addressing.

Thus, for our purposes, it's always CIDv1.

PR #835, opened 2025-05-20
fix: remove REDIS_OPTIONS from docker compose files

JoeCap08055:
# Description
Since `REDIS_URL` and `REDIS_OPTIONS` are now mutually exclusive, make sure any docker compose files do not specify both.

### DIFF
--- a/docker-compose-e2e.account.yaml
+++ b/docker-compose-e2e.account.yaml
@@ -2,7 +2,6 @@
 x-common-environment: &common-environment
   FREQUENCY_API_WS_URL: ${FREQUENCY_API_WS_URL:-ws://frequency:9944}
   SIWF_NODE_RPC_URL: ${SIWF_NODE_RPC_URL:-http://localhost:9944}
-  REDIS_OPTIONS: ${REDIS_OPTIONS:-""}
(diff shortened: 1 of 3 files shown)


shannonwells (commented):
Is there any reason not to get rid of it everywhere?  It's also in values.yml, docker-compose-e2e.content-publishing.yml and docker-compose.swarm.yaml

JoeCap08055:
> Is there any reason not to get rid of it everywhere? It's also in values.yml, docker-compose-e2e.content-publishing.yml and docker-compose.swarm.yaml

Not sure what the correct thing to do is. I originally added REDIS_OPTIONS as supplemental to REDIS_URL, before I realized that all options can be specified in the URL itself. I then left REDIS_OPTIONS in to give the option of specifying the Redis connection via a JSON object or a URL... but maybe it's best if we just stick with one method. We could instead keep REDIS_OPTIONS as *only* supplemental to REDIS_URL, but mark it deprecated...

shannonwells (approved):
The previous failure is fixed with this change, I say ship it. Though it does no harm to remove it elsewhere, it seems; I have changes like that locally.

PR #837, opened 2025-05-28
Fix issue with ipfs.files.stat

wesbiggs:
# Problem

`ipfs.files.stat` was always returning an error: `paths must start with a leading slash` instead of a `FilesStatResult`.

In addition, the tests weren't correctly mocking `ipfs.files.stat` (they were still referencing `ipfs.block.stat`).

# Solution

Pass ipfs.files.stat a path, not a CID.

Also correct mocks and tests.

## Steps to Verify:

See if batches get announced successfully.

### DIFF
--- a/__mocks__/kubo-rpc-client.ts
+++ b/__mocks__/kubo-rpc-client.ts
@@ -27,7 +27,7 @@ export class CID {
   }
 }
 
-export interface BlockStatResult {
+export interface FilesStatResult {
   cid: string;
   size: number;
 }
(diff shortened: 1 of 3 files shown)


JoeCap08055:
@wesbiggs I added a small CI pipeline refactor to this PR to make sure the tests you fixed are actually run in CI when appropriate, and also to refactor some redundant pipeline steps that only need to be run once, not per-service.

wesbiggs:
Good catch, thanks @JoeCap08055!

shannonwells (approved):
I don't have enough context to evaluate this PR but @JoeCap08055 approved and `nest build libs` worked for me, so approving

Issue #315, opened 2024-07-30
Content Publishing Service Success Webhook


wilwade:
### Context
When integrating with the Content Publisher, I want to be able to know for sure that a piece of content has successfully met these criteria:

1. The content is published to IPFS 
2. The content is in a batch
3. That batch is in a finalized block

#### How to handle error cases?
When content has an error, it should error out.

### Any other comments?

This should be as close to Graph Service webhook as possible.

### Success Payload

- Id
- Block number (if success)
- Status
- ~Message Content~ (Maybe future)

### Railroad

1. Submit content to content publisher (Respond with 201 and Id)
3. Perform validations, Error: `Webhook(error)`
4. Submit to chain, Error: `Webhook(error)`
5. Wait for finalized block, Mortality/RPC Error: `Webhook(error)`
6. In finalized block `Webhook(success)`

shannonwells:
Partially addressed in #1039

PR #842, opened 2025-06-06
Upload batch announcement

### LINKED ISSUE #727: [content-publishing] batchAnnouncement endpoint should include asset upload
# Description
Currently, as developed for the "freesky" PoC, the `POST /v2/content/batchAnnouncement` endpoint requires a user to have first uploaded asset files to IPFS using the `PUT /v1/asset/upload` endpoint. This approach has several issues:

1. The `asset/upload` endpoint does extra processing to cache asset meta-information in Redis that is not necessary for pre-built Parquet batch files
2. Attempting to create a batch announcement without having previously uploaded a batch is an error scenario, while uploading a batch file and never submitting a batch announcement is simply a waste of resources; therefore, it's better to integrate the two steps into a single endpoint.
3. A dedicated upload for batch files could more easily do additional validation of file type & schema

# Acceptance Criteria
- [ ] `POST /v2/content/batchAnnouncement` endpoint changed to a `/v3 PUT` endpoint that includes the Parquet files themselves, instead of the CID references of a previously-uploaded file
- [ ] (OPTIONAL) Add validation of file type (Parquet, others?) against the requested schema (ie, validate file against actual Parquet schema)

### LINKED ISSUE #822: [content-publishing] Reduce capacity use by batching multiple chain submissions to the chain using payWithCapacityBatchAll
Currently each batch published to the chain is a one off action.

However, the chain has `payWithCapacityBatchAll` which would reduce the amount of transactions that are needed and thus lower the cost in capacity when publishing multiple batches at the same time.

As a provider I want to have the content publishing use less capacity

emresurmeli:
# Problem
Closes #727 
Closes #822 

We want to combine `v2/asset/upload` with `v2/content/batchAnnouncement` to reduce file uploading to a single endpoint. We also want to modify the existing batch announcement capability to write batches to the chain using a single `payWithCapacityBatchAll()` call rather than many `payWithCapacity()` calls. 

# Solution
This draft creates an endpoint POST `v3/content/batchAnnoucement` which combines uploading files to IPFS, and enqueuing the batch to the `batchAnnouncerQueue`. 

Batch announcements are then batched up in the MessagePublisher and written to the chain using `payWithCapacityBatchAll()` extrinsic.

# Testing
You will want to run the following docker containers:
`content-publishing-service-api`
`content-publishing-service-watcher`
`redis`
`ipfs`

You can use a HTTP client to test the endpoint, I'm using Postman
Postman HTTP POST

Check the BullMQ queues for successful processing of jobs through the batchQueue and publishQueue
You can access the BullMQ UI @ http://localhost:3010/queues/
BullMQ UI

Confirm successful processing of file uploads and batchAnnoucements through the content publishing service API logs
API logs on the container

Check to see the jobs are successfully being picked up and processed by the workers
Worker logs on the container

And finally confirm the transaction was included in a block via PolkadotJS UI
PolkadotJS UI

# To do
- [x] Docs  
- [x] Testing edge cases (Especially of batching things)
- [x] Finishing e2e testing
- [x] Pair with a senior to sanity check the flow of this new endpoint


apps/content-publishing-api/src/controllers/v3/content.controller.v3.ts
```diff
@@ -0,0 +1,138 @@
...
+    summary: 'Upload files and create batch announcements',
+    description:
+      'Upload files with their corresponding schema IDs to create batch announcements. Each file must be accompanied by a schemaId field.',
+  })
+  @ApiBody({
+    description: 'Asset files',
+    type: FilesUploadDto,
+  })
+  @ApiResponse({ status: '2XX', type: BatchAnnouncementResponseDto })
+  async postUploadBatchAnnouncement(
```
JoeCap08055, line 48: Q: since the file upload logic is the same as `POST /v2/asset/upload` (a great example of NOT following the above-mentioned REST endpoint-naming pattern--but we chose consistency with v1, which was implemented before we started following that standard), I wonder if any of the file upload handling could be refactored to a common function?

apps/content-publishing-api/src/controllers/v3/content.controller.v3.ts
```diff
@@ -0,0 +1,138 @@
...
+        const errors = uploadResults.filter((result) => result.error);
+        if (errors.length > 0) {
+          throw new BadRequestException(errors[0].error || 'File upload failed');
+        }
+
+        // Process each file and schema ID pair
+        const batchPromises = uploadResults.map((result, index) => {
+          if (!result.cid) {
+            throw new Error('Missing CID in upload result');
+          }
```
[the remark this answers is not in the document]
JoeCap08055: At this point in the process, we've successfully uploaded all of the files--so throwing here would mean the client has no idea the files were uploaded, and would therefore retry the whole process. Whereas, if we return the uploaded file identifiers and whether each one was successfully enqueued as a batch announcement task, then any failures could be retried using the `POST /v2/content/batchAnnouncement` endpoint to announce an already-uploaded batch file.

apps/content-publishing-api/src/controllers/v3/content.controller.v3.ts
```diff
@@ -0,0 +1,138 @@
...
+            cid: result.cid,
+            schemaId: schemaIds[index],
+          });
+        });
+
+        // Submit all batch requests and collect responses
+        const responses = await Promise.all(batchPromises);
+
+        // Get the first reference ID as the batch reference ID
+        const batchReferenceId = responses[0].referenceId;
```
JoeCap08055, line 111: We probably want to return the individual batch reference IDs. This will enable us to:
1. Know whether each file uploaded was successfully enqueued as a batch announcement task
2. Track the status of each batch announcement task (functionality to be built, see #315 )

JoeCap08055:
Regarding this approach: we've had some issues so far with handling all edge cases when processing the request stream directly via Busboy. IPFS on the other end seems to be somewhat fragile, and often errors on the upload, which causes us issues in our stream processing.

I had been thinking of an alternate approach--probably the subject of a separate issue.
The main reasons we switched from using Multer to using Busboy directly were:
* Multer, by default, caches all of the files in app heap memory. This is not ideal.
    * The other standard option requires configuring local disk storage to hold the files, whether permanently or temporarily
* We wanted to eliminate the path we had been taking, which was to store the entire file contents in Redis once they were received. Not only is this a bad use case for Redis in general, but there is specifically a hard 512MB limit when doing so (Redis-internal)
* Because adding the files to be uploaded to a task queue & returning the CID did not give us visibility into the completion status of the upload, we ended up with some race conditions in `content-publishing`. The normal workflow is (1) upload one or more assets, (2) create an announcement referencing those assets and request it to be published on-chain. This was happening so fast and with such volume that the announcement came in before the assets had been uploaded, so the task would fail when it checked for the presence of the assets. Having the upload endpoint actually stream the upload to the final destination meant that the client would know for certain that the assets had been successfully uploaded to IPFS when they got the response.

I had a chat recently with @shannonwells in which I suggested we might combine the two approaches for something a little more fault-tolerant. We can still stream the upload direct to IPFS, but instead of hand-rolling our own Busboy code, we could write a [custom storage engine](https://github.com/expressjs/multer/blob/master/StorageEngine.md) for the Multer middleware that would handle streaming to IPFS and modify the multipart request to contain the resulting CID. This way, we'd still avoid caching the files in memory or Redis, but the controller endpoint itself would just use the Multer middleware like the `/v1/asset/upload` endpoint, which is much simpler to deal with.

Perhaps this ☝🏻 should be captured in its own issue.

libs/storage/src/ipfs/ipfs.service.ts
```diff
@@ -79,14 +79,24 @@ export class IpfsService {
...
+        method: "HEAD"
+      });
+      
+      if (response.ok) {
+        return true;
+      }
+
+      // If not found in gateway, check if we can get it from the IPFS network
+      const stat = await this.ipfs.block.stat(CID.parse(cid));
+      return !!stat.cid;
```
JoeCap08055, line 95: This seems to contradict the purpose indicated by the function name

apps/content-publishing-api/test/app.e2e-spec.ts
```diff
@@ -828,11 +828,80 @@ describe('AppController E2E request verification!', () => {
...
+      const files = Array(config.fileUploadCountLimit + 1).fill(Buffer.from('fake image content'));
+      const req = files.reduce(
+        (acc, file, index) =>
+          acc
+            .attach('files', file, { filename: `test${index}.jpg`, contentType: 'image/jpeg' })
+            .field('schemaId', '16001'),
+        request(app.getHttpServer()).post('/v3/content/batchAnnouncement'),
+      );
+
+      await req.expect(400).expect((res) => expect(res.text).toContain('Max file upload count'));
```
shannonwells, line 862: Can this be changed to match the full error message?
```suggestion
      await req.expect(400).expect((res) => expect(res.text).toContain('Max file upload count per request exceeded'));
```

apps/content-publishing-worker/src/publisher/message.publisher.ts
```diff
@@ -10,27 +9,63 @@ import { DelayedError } from 'bullmq';
...
+    if (this.messageQueue.length >= this.BATCH_SIZE_LIMIT) {
+      return this.processBatch();
+    }
+
+    // If this is the first message in the queue, process it immediately
+    if (this.messageQueue.length === 1) {
+      return this.processBatch();
+    }
+
+    // Otherwise, wait for more messages to accumulate
```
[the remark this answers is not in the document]
emresurmeli: This wasn't supposed to be left in there

apps/content-publishing-worker/src/batch_announcer/batch.announcer.ts
```diff
@@ -24,7 +22,7 @@ export class BatchAnnouncer {
     private blockchainService: BlockchainService,
     private ipfsService: IpfsService,
   ) {
-    this.logger = pino(getBasicPinoOptions(BatchAnnouncer.name));
+    this.logger = new Logger(BatchAnnouncer.name);
```
shannonwells, line 25: Why is this going back to the native logger?

retrieved_file.jpg
shannonwells, line 1: It's better to create test files and put them in /tmp, but if you must check in a test binary file, it should be smaller and not stored at the top level, but in some test file directory.  Also I should think you don't need more than one.

openapi-specs/content-publishing.openapi.json
```diff
@@ -541,6 +541,89 @@
...
+                            "type": "string",
+                            "description": "Unique identifier for tracking the batch announcement"
+                          },
+                          "cid": {
+                            "type": "string",
+                            "description": "IPFS CID of the uploaded file"
+                          },
+                          "error": {
+                            "type": "string",
+                            "description": "Error message if the file upload or batch creation failed"
```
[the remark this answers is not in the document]
JoeCap08055: It's a bit more complex, I think. The endpoint accepts multiple files for upload; hence the response contains an array of statuses. We may get a `2xx` response from the endpoint, but it's still incumbent on the client to check the individual upload status of each file.

I suppose additional logic could be applied, wherein if ALL files in the request have an error status, then the response could be `4xx` or `5xx`.

apps/content-publishing-api/src/controllers/v3/content.controller.v3.ts
```diff
@@ -0,0 +1,143 @@
...
+      if (fileIndex > this.appConfig.fileUploadCountLimit) {
+        fileStream.resume(); // Make sure we consume the entire file stream
+        throw new BadRequestException('Max file upload count per request exceeded');
+      }
+
+      fileProcessingPromises.push(
+        this.apiService.uploadStreamedAsset(fileStream, fileinfo.filename, fileinfo.mimeType),
+      );
+    });
+
```
JoeCap08055, line 81: From this point forward, if we throw without forming a valid response, we risk losing information:
* At this point, all files uploads have been started
* If we subsequently throw an error, we could lose information about any successfully uploaded assets
* If we use `Promise.all` vs `Promise.allSettled`, we also risk throwing and losing information

I suggest that _almost_ any error from this point forward still generates a valid response, containing the success or error information of each file in the request. Exceptions to this would be:
* Only one file, and it errors on upload
* Multiple files, ALL error on upload
* Bad Request -- ie, zero files or mismatched schema/file counts

Batch request submissions should never result in throwing an error, because at that point the corresponding file has already successfully been uploaded.

apps/content-publishing-api/src/controllers/v3/content.controller.v3.ts
```diff
@@ -0,0 +1,143 @@
...
+          return this.apiService
+            .enqueueBatchRequest({
+              cid: uploadResult.cid,
+              schemaId: schemaIds[uploadResults.indexOf(uploadResult)],
+            })
+            .catch(() => null);
+        });
+
+        // Submit all batch requests and collect responses
+        const batchResults = await Promise.all(batchPromises);
```
JoeCap08055, line 116: This `Promise.all` is okay because we map rejections to `null` with a `.catch(() => null)` above

emresurmeli:
I've modified the controller to handle any upload issues by requiring a retry and made changes to the message publishing queue to prevent it from being cleared before successful processing. I also added some decorators to the request DTO and updated the OpenAPI spec JSON to reflect the changes based on your feedback @wilwade @JoeCap08055 

One thing to note, Joe, is that `Promise.all` for uploads can throw an `Error: "Failed to process batch announcement"`, but if we want to, we could also ensure that we inform the user which file had an upload error. Could be useful 🤔

JoeCap08055:
> I've modified the controller to handle any upload issues by requiring a retry and made changes to the message publishing queue to prevent it from being cleared before successful processing. I also added some decorators to the request DTO and updated the OpenAPI spec JSON to reflect the changes based on your feedback @wilwade @JoeCap08055

See [my other note](https://github.com/ProjectLibertyLabs/gateway/pull/842#discussion_r2251707796)

> One thing to note, Joe, is that `Promise.all` for uploads can throw an `Error: "Failed to process batch announcement"`, but if we want to, we could also ensure that we inform the user which file had an upload error. Could be useful 🤔

Could use `Promise.allSettled` instead of `Promise.all`; this would allow all of the async Promises to run to completion and return an array of statuses for each one. With `Promise.all`, you have no idea which ones resolved; you only know that at least one of them rejected.

apps/content-publishing-api/src/controllers/v3/content.controller.v3.ts
```diff
@@ -84,45 +109,82 @@ export class ContentControllerV3 {
...
+            responseFiles.push({ error });
+          }
+        });
 
-        // Check for upload errors
-        const errors = uploadResults.filter((uploadResult) => uploadResult.error);
-        if (errors.length > 0) {
-          throw new BadRequestException(errors[0].error || 'File upload failed');
+        // Create batch announcements for successful uploads
+        if (successfulUploads.length > 0) {
```
JoeCap08055, line 138: Discussion point: is it better to take an all/nothing approach here; ie, if there are any upload failures, skip the batch announcements completely? I don't think I have a strong opinion here either way, but probably worth discussing.

apps/content-publishing-api/src/controllers/v3/content.controller.v3.ts
```diff
@@ -84,45 +109,82 @@ export class ContentControllerV3 {
...
+          const batchPromises = successfulUploads.map(({ uploadResult, originalIndex }) =>
+            this.apiService.enqueueBatchRequest({
+              cid: uploadResult.cid!,
+              schemaId: schemaIds[originalIndex],
+            }),
+          );
+
+          try {
+            const batchResults = await Promise.all(batchPromises);
```
JoeCap08055, line 147: Should probably use `Promise.allSettled` here as well

apps/content-publishing-api/src/controllers/v3/content.controller.v3.ts
```diff
@@ -84,45 +109,82 @@ export class ContentControllerV3 {
...
+            responseFiles.push({ cid: uploadResult.value.cid });
+          } else {
+            hasFailedUploads = true;
+            const error =
+              uploadResult.status === 'rejected'
+                ? uploadResult.reason?.message || 'Upload failed'
+                : uploadResult.value.error || 'Upload failed';
+            responseFiles.push({ error });
+          }
+        });
```
JoeCap08055, line 135: code golf; I like destructuring 😄 
just a suggestion; take it or leave it at your preference
```suggestion
        uploadResults.forEach(({ status, reason, value: uploadResult }, index) => {
          if (status === 'fulfilled' && !!uploadResult.cid && !uploadResult.error) {
            successfulUploads.push({ uploadResult, originalIndex: index });
            // Placeholder - will be filled after batch creation
            responseFiles.push({ cid: uploadResult.cid });
          } else {
            hasFailedUploads = true;
            responseFiles.push({ error: reason?.message || uploadResult?.error || 'Upload failed' });
          }
        });
```

apps/content-publishing-api/src/controllers/v3/content.controller.v3.ts
shannonwells: Agree that respond with upload status of each item in the batch.  I'm not sure of the value of skipping batch announcements entirely if some are successful.  As long as the reponse can be parsed in order to try again and debug.

libs/logger/load-test-worker-logs.json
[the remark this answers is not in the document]
emresurmeli: Ah yes

apps/content-publishing-api/src/controllers/v3/content.controller.v3.ts
shannonwells: Let's table this for now and have an in-person whiteboarding discussion on it to decide.

shannonwells (approved):
Let's go ahead and get this merged (with the 1-2 changes we discussed in the review), and do a couple of TBD follow-ups.

Issue #477, opened 2024-09-03
Monitoring and Logging Tooling


p5150j:
As a system administrator,
I want to integrate monitoring and logging tools like Grafana, Prometheus, and CloudWatch, so that I can monitor the health and performance of the Gateway stack and receive real-time alerts on any issues.  

We want to be able to reduce the lift for DevOps whenever there is an issue instead of making them go fetch logs and send them to us.

Acceptance Criteria:

- [x] **HIGH PRIORITY** Gateway services use a JSON-enabled logger (ie, pino, winston, etc). Pino and Winston are loggers, Pinot has been used in past.  Another possibility is upgrade to NestJS 11 which has JSON-enabled logger, but we don't know if it's robust/usable enough for what we want.
- [x] Enable Prometheus endpoints:  Prometheus should collect and store metrics, while CloudWatch should be configured for logging and alerting.  See requests for what pertinent data goes into the metrics below
    - [x] accounts-api
    - [x] accounts-worker
    - [x] graph-api
    - [x] graph-worker
    - [x] content-publisher-api
    - [x] content-publisher-worker
    - [x] content-watcher*

- README Documentation on how to access and interpret monitoring and logging data should be provided.
- [x] Documentation should be updated with specifics on how to set log levels, e.g. `DEBUG=true`, `VERBOSE_LOGGING=true`.
- [x] #734

**What's the ask for the Gateway team here?  Make better logs?**
- Monitoring and logging tools (Grafana, Prometheus, CloudWatch) should be integrated with the Gateway stack.
- Dashboards should be set up in Grafana to display key performance metrics and system health.
- Alerts should be set up for key events (e.g., high CPU usage, errors, etc.) to notify the system administrator in real-time.  May need to include relevant errors in Prometheus

Note:
Development teams need to add and relate all necessary implementation subtasks (e.g., integrating each monitoring tool, setting up dashboards and alerts, configuring data collection) to this story for detailed execution and tracking.

shannonwells:
Completed with #897

PR #843, opened 2025-06-09
Feat/logging and prometheus

shannonwells:
# Problem

Instrument Gateway services for JSON output in the logs, and also a prometheus endpoint *for accounts-api only*, the rest will come in a follow-up. #477 

# Solution
* Instrument with pino using `nestjs-pino`.  Instrument for prometheus using `prom-client`
* Separate log level from formatting by using `LOG_LEVEL` and `PRETTY` for environment variables.

## NOTES
* this turned out to be a bigger task than assumed, because of the different way NestJS works with loggers in different modules, and in tests. 

### Change summary
* global change of logger.log -> logger.info
* global change of logger.verbose -> logger.trace
* Config getters that everybody uses in `libs/logger/logLevel-common-config.ts`
* Each of these has a _different_ way to instantiate loggers:
  * Apps (in main.ts)
  * controllers
  * services
  * stuff outside of NestJS which just uses bare pino
  * sometimes components are special snowflakes and just have to do things differently, like BaseConsumer which instantiates the logger for you, and BlockchainScannerService which expects an already initialized logger.

## Steps to Verify:
1. npm install
1. copy new environment templates and adjust `LOG_LEVEL` to one of error, warn, info, debug, trace.  Adjust `PRETTY=true` or `false`.
1. run gateway using the method of your choice - can also just run one of the services, like accounts.
1. verify that your log settings are respected in the services you made settings for.
1. `curl https://localhost:3000/metrics`  (Use port 3013 if running in docker)
1. verify a 200 status + some cpu and memory data
1. verify basic endpoints work, like `GET acccounts/[:msa_id]` either by curl or swagger docs - make sure there are no 500 internal server errors.
1. CI is passing, obvy


.eslintrc.js
```diff
@@ -54,7 +54,7 @@ module.exports = {
     'import/order': 'off',
     'no-await-in-loop': 'off',
     'no-useless-constructor': 'off',
-    'no-empty-function': 'warn',
+    'no-empty-function': 'off',
```
shannonwells, line 57: this was requiring a lot of `eslint-disable-next-lines` and just seemed like we don't really need this on.

apps/account-api/src/controllers/v1/delegation-v1.controller.ts
```diff
@@ -57,7 +57,7 @@ export class DelegationControllerV1 {
     @Param() { accountId }: AccountIdDto,
     @Param() { providerId }: ProviderMsaIdDto,
   ): Promise<RevokeDelegationPayloadResponseDto> {
-    this.logger.verbose(`Getting RevokeDelegationPayload for account ${accountId} and provider ${providerId}`);
+    this.logger.trace(`Getting RevokeDelegationPayload for account ${accountId} and provider ${providerId}`);
```
shannonwells, line 60: `verbose` is now `trace` with Pino.

apps/account-api/src/controllers/prometheus.controller.ts
```diff
@@ -0,0 +1,20 @@
+import { Controller, Get, HttpCode, HttpStatus, Header } from '@nestjs/common';
+import { ApiOkResponse, ApiOperation, ApiTags } from '@nestjs/swagger';
+import { register, collectDefaultMetrics } from 'prom-client';
+
+collectDefaultMetrics();
+
+@Controller()
+@ApiTags('metrics')
+export class PrometheusController {
```
shannonwells, line 9: The metrics endpoint.

apps/account-api/src/controllers/v1/handles-v1.controller.ts
```diff
@@ -77,7 +74,7 @@ export class HandlesControllerV1 {
       ...createHandleRequest,
       type: TransactionType.CREATE_HANDLE,
     });
-    this.logger.log(`createHandle in progress. referenceId: ${response.referenceId}`);
+    this.logger.info(`createHandle in progress. referenceId: ${response.referenceId}`);
```
shannonwells, line 77: `log` becomes `info` with Pino.

apps/account-api/src/main.ts
```diff
@@ -37,10 +39,11 @@ async function bootstrap() {
   });
 
   const app = await NestFactory.create<NestExpressApplication>(ApiModule, {
-    logger: getLogLevels(),
     rawBody: true,
   });
 
+  app.useLogger(app.get(PinoLogger));
```
shannonwells, line 45: This `useLogger` call is only ever used when constructing an app, so only in `main.ts` files.

apps/account-api/src/services/siwfV2.service.spec.ts
```diff
@@ -51,7 +52,7 @@ const mockAccountApiConfigProvider = GenerateMockConfigProvider<IAccountApiConfi
   graphEnvironmentType: 0,
   siwfUrl: '',
   siwfV2Url: 'https://www.example.com/siwf',
-  siwfV2URIValidation: 'localhost',
+  siwfV2URIValidation: ['localhost'],
```
shannonwells, line 55: eslint was complaining about this

apps/account-api/src/services/siwfV2.service.spec.ts
```diff
@@ -227,7 +229,7 @@ describe('SiwfV2Service', () => {
     it('Should throw BadRequest if the response is for a different domain', async () => {
       jest
         .spyOn(mockAccountApiConfigProvider.useValue, 'siwfV2URIValidation', 'get')
-        .mockReturnValue('bad.example.com');
+        .mockReturnValue(['bad.example.com']);
```
shannonwells, line 232: same here.

libs/logger/logLevel-common-config.ts
```diff
@@ -1,23 +1,67 @@
...
-
 /**
  * Determines appropriate log levels based on environment variables
  * - Always includes 'error', 'warn', and 'log'
  * - Adds 'debug' when DEBUG env var is present
  * - Adds 'verbose' when VERBOSE_LOGGING is true or DEBUG=verbose
  */
-export function getLogLevels(): LogLevel[] {
-  const logLevels: LogLevel[] = ['error', 'warn', 'log'];
+export function getLogLevels(): string[] {
```
shannonwells, line 7: may not need this any more.

libs/logger/logLevel-common-config.ts
```diff
@@ -1,23 +1,57 @@
...
+}
 
-  // Add verbose logging if specifically requested
-  if (process.env.VERBOSE_LOGGING || process.env.DEBUG === 'verbose') {
-    logLevels.push('verbose');
+export function getPinoTransport() {
+  if (process.env.PRETTY === 'true') {
+    return {
+      target: 'pino-pretty',
+      options: {
```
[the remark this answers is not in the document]
emresurmeli: Looked pretty good to me

apps/account-api/src/controllers/prometheus.controller.ts
emresurmeli: Wish it returned JSON! Looks like there might be a method to do that if we want `register.getMetricsAsJSON()`

emresurmeli (approved):
I tested a few endpoints and they look great! The only thing that comes to mind is maybe making the logger its own service, where we can pass it params, since we have to use it in so many different ways depending on where we want to instantiate it. 

But I think that could be another issue.

apps/account-api/src/api.module.ts
```diff
@@ -23,6 +25,7 @@ import apiConfig from './api.config';
 import { APP_FILTER } from '@nestjs/core';
 import { AllExceptionsFilter } from '#utils/filters/exceptions.filter';
 import { QueueModule } from '#queue/queue.module';
+import { getPinoHttpOptions } from '../../../libs/logger/logLevel-common-config';
```
JoeCap08055, line 28: nit: it would be nice to create a module alias like `#logger`, like we did for all of the other modules. Relative-path imports like this are ugly, IMO

apps/account-api/src/services/accounts.service.ts
```diff
@@ -31,7 +33,7 @@ export class AccountsService {
     private blockchainService: BlockchainRpcQueryService,
     private enqueueService: EnqueueService,
   ) {
-    this.logger = new Logger(this.constructor.name);
+    this.logger = pino(getBasicPinoOptions(AccountsService.name));
```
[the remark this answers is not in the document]
JoeCap08055: Here be dragons!

I may be misremembering, but there are 2 ways to instantiate a Pino logger. The way you're doing it here gives you a "100% Pino" logger, with all of Pino's API. However, swapping the logger this way may require examining individual log calls, because Pino doesn't marshal its arguments the same way the Nest logger does.

On the other hand, since you've instantiated the app and set the 'master' logger to the Pino logger using the Nestjs-wrapped Pino logger from the `nestjs-pino` package, if you simply instantiate a normal "NestJS" logger via `new Logger(...)`, it will use the Pino logger under the hood, but continue to present the NestJS Logger API.

This makes a significant difference when log calls include additional arguments beyond just a single string.

apps/account-api/src/api.module.ts
shannonwells: done

PR #846, opened 2025-06-10
Revert the redis.config file

emresurmeli:
New configs are messing up tests - revert to previous version


shannonwells (approved):
Do it to it

PR #849, opened 2025-06-12
logging in pino just for account- apps

shannonwells:
# Problem

problem statement, including
Link to GitHub Issue(s):

# Solution

What I/we did to solve this problem

with @pairperson1

## Steps to Verify:

1. What to do next
1. Any other instructions
1. Expected behavior
1. Suggestions for testing

## Screenshots (optional):


libs/logger/src/logLevel-common-config.ts
```diff
@@ -0,0 +1,75 @@
+import { LogLevel } from '@nestjs/common';
+
+/**
+ * Determines appropriate log levels based on environment variables
+ * - Always includes 'error', 'warn', and 'log'
+ * - Adds 'debug' when DEBUG env var is present
+ * - Adds 'verbose' when VERBOSE_LOGGING is true or DEBUG=verbose
```
JoeCap08055, line 7: This comment no longer matches the code.

I assume this is only here to support the apps that haven't been converted to pino yet.
It may be worth considering whether we want to support this style of behavior with pino logging, ie:

Default Pino behavior: ordered levels `fatal, error, warn, info, debug, trace, silent`
Our previous behavior allowed (translating level names) enabling *either* debug, or trace, or both. Is that something worth doing? Could be done with custom log levels in pino. Just a thought.

shannonwells: I should have noted, this will be removed with the final PR.

JoeCap08055: I figured as much--any thoughts on standard log behavior vs custom levels?

The issue was that we wanted to be able to enable 'trace' without enabling 'debug'. But maybe that was the wrong approach? I think the issue was that 'debug' enables a lot of noisy structures, whereas 'trace' enables a lot of messages, but less data?

I guess there's a reason for the standard; I'm fine with anything as long as it's consistently applied. Might be worth taking a spin through any `.verbose()` or `.trace()` messages to see what's there.

All to say, this is non-blocking for this PR, especially since there are more coming.

JoeCap08055 (approved):
One comment--once all apps have transitioned, we'll need a final review of all log messages. Any log calls with > 1 argument need to be examined to make sure they conform to the Pino way of logging.

apps/account-api/src/main.ts
```diff
@@ -1,15 +1,18 @@
...
 import { ApiModule } from './api.module';
 import { TimeoutInterceptor } from '#utils/interceptors/timeout.interceptor';
 import { NestExpressApplication } from '@nestjs/platform-express';
 import apiConfig, { IAccountApiConfig } from './api.config';
 import { generateSwaggerDoc, initializeSwaggerUI, writeOpenApiFile } from '#openapi/openapi';
-import { getLogLevels } from 'libs/logger/logLevel-common-config';
+import { getBasicPinoOptions, getCurrentLogLevel } from '#logger-lib';
 
-const logger = new Logger('main');
+import { Logger as PinoLogger } from 'nestjs-pino';
```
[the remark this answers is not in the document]
emresurmeli: Ah, my fault - I was skimming and has this as Logger, PinoLogger for some reason :face_palm:

PR #860, opened 2025-06-19
Logger cleanup

shannonwells:
Change how logger.error is called for better stack traces. 
Set up some remaining services to use Pino.

# Problem
Relates to #477

### DIFF
--- a/apps/account-api/src/main.ts
+++ b/apps/account-api/src/main.ts
@@ -33,17 +33,11 @@ function startShutdownTimer() {
 }
 
 async function bootstrap() {
-  process.on('uncaughtException', (error) => {
-    console.error('****** UNCAUGHT EXCEPTION ******', error);
-    process.exit(1);
-  });
-
   const app = await NestFactory.create<NestExpressApplication>(ApiModule, {
     rawBody: true,
   });
 
   app.useLogger(app.get(PinoLogger));
-
   // Enable URL-based API versioning
   app.enableVersioning({
     type: VersioningType.URI,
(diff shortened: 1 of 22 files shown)


shannonwells:
@emresurmeli NestJS really wants to use its own logger until after a bunch of other stuff has initialized.  There's a way to force it, but since the only remaining messages are for NestJS's `NestFactory.create` and `InstanceLoader`, I figured it should be fine to leave it.  Everything else is using the new loggers now.

apps/account-api/src/main.ts
```diff
@@ -33,17 +33,11 @@ function startShutdownTimer() {
 }
 
 async function bootstrap() {
-  process.on('uncaughtException', (error) => {
```
shannonwells, line 36: I'm unsure why this is here; it's not in all the apps main.ts, and when I was testing error output using just a random throw call, this never gets tripped because of the try/catch at the end.  :woman_shrugging:

apps/content-publishing-api/src/api.service.ts
```diff
@@ -327,7 +327,7 @@ export class ApiService {
         calculateIncrementalDsnpMultiHash(hashPassThru),
       ]);
     } catch (error: any) {
-      this.logger.error('❌ Upload/hash promise error:', error.message);
+      this.logger.error(`❌ Upload/hash promise error:, ${error.message}`);
```
hwahby, line 330: Nit: Makes sense to turn it into a template string if we want a single string. A little curious why the difference in how we call these. Do we care about only the message vs the whole error object?

this.logger.error(err, '❌ Stream error:')
vs
this.logger.error(`❌ Upload/hash promise error:, ${error.message}`);

PR #862, opened 2025-06-24
prometheus endpoints, README entries

shannonwells:
# Problem
Add prometheus endpoints to all app apis. 

I chose _not_ to add prometheus to the workers, because they are microservices which currently do not expose any endpoints. To add this to workers requires some thought about how  best to set this up (maybe a sidecar service) and right now we have higher priorities.  I've created issue #864 for this task.

Part of #477 

# Solution
- [x] Use nestjs-prometheus module to stand up a `/metrics` endpoint with the default output.
- [x] Add entries for the logging settings and prometheus endpoints to README.
- [x] Customize config to show app name

## Steps to Verify:

1. Run `./start.sh`, accept defaults.
1. Navigate to each of the 4 metrics endpoints for account-api, content-publishing-api, content-watcher, and graph-api, and verify that `/metrics` endpoints work on all 4 apps, i.e.
  * http://localhost:3010/metrics,
  * http://localhost:3011/metrics,
  * http://localhost:3012/metrics,
  * http://localhost:3013/metrics,

If you like, you can [install prometheus](https://prometheus.io/docs/prometheus/latest/installation/) locally, point it at the 4 services, let the apps run for a bit and then look at the pretty graphs. To see something like below, edit `prometheus.yml` and under `static_configs`, set:
```
  - targets: ["localhost:3000", "localhost:3013", "localhost:3012", "localhost:3011", "localhost:3010"]
```

Screenshot from 2025-06-24 12-29-39

### DIFF
--- a/.spellcheckerdict.txt
+++ b/.spellcheckerdict.txt
@@ -23,6 +23,7 @@ codebase
 ConfigMap
 content-publishing-service-api
 CORS
+Ctrl-C
 cryptographic
 DaemonSet
 dapps
@@ -37,6 +38,7 @@ DTO
 E2E
 EC2
 (enqueueing|enqueuing)
+.env
 enum
 etcd
 extrinsics
@@ -81,6 +83,7 @@ mjs
 [Mm]icroservices?(-based)?
 Middleware
 MSA
+MSAs
 namespace
 NestJS
 (NGINX|Nginx)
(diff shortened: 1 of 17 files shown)


hwahby:
This looks good. I like the additions to the documentation. 
I confirmed that I could make a request against an endpoint and view it's data on the dashboard.

A follow-up might to expand to the worker services and ingest Redis/queue metrics into prometheus.

PR #870, opened 2025-06-26
use a real key in graphKey examples

### LINKED ISSUE #861: The Example Response for the Account Service `/v2/accounts/siwf` Response has a bad `graphKey` value
Mistake is in two places:
- https://projectlibertylabs.github.io/gateway/account/#tag/v2accounts/operation/AccountsControllerV2_postSignInWithFrequency
- https://projectlibertylabs.github.io/gateway/GettingStarted/SSO.html#step-3-handle-the-callback

The example has `"graphKey": "555-867-5309",` which doesn't even match the type of the response which is `GraphKeySubject` type.

Expected: A reasonable `GraphKeySubject` type example

shannonwells:
# Problem
The graphKey example in the docs is a phone number and that's wrong.

Closes #861 

# Solution
Use an actual Frequency-compatible test public key as an example.


docs/src/GettingStarted/SSO.md
```diff
@@ -1,7 +1,8 @@
 # Fast Single Sign On with SIWF v2 with Account Service
 
 Sign In With Frequency (SIWF v2) is quick, user-friendly, decentralized authentication using the Frequency blockchain.
-Coupled with the Account Service, this provides fast and secure SSO across applications by utilizing cryptographic signatures to verify user identities without complex identity management systems.
+Coupled with the Account Service, this provides fast and secure SSO across applications by utilizing cryptographic
```
shannonwells, line 4: linter fights, nobody wins.

emresurmeli (approved):
Yeah lgtm - I don't know how to confirm this graph key is real so I am just approving the changes with no way to test on my end :)

Issue #867, opened 2025-06-25
Gateway needs to expose a getBlocknumber call for signing


shannonwells:
The embedded wallet api needs to be able to calculate an expiration block rather than some number in the future, but gateway doesn't expose any such endpoint. The payload has to be signed, so it has to have the actual expiration block; may as well allow passthrough query because there would have to be a round trip anyway.

## endpoint spec decided:
```
javascript /v1/frequency/blockInfo
```
returns:
```javascript
blocknumber: number;
finalized_blocknumber: number
genesis: "0x..."; // string
runtime_version: number; 
```

## Acceptance criteria
- [ ] Add a passthrough endpoint to account-api that calls blockchain-rpc-query.service's `getBlock` function and return the above structure.
- [ ] Add the same to graph-api
- [ ] tests for above

wilwade:
@shannonwells / @hwahby / @emresurmeli  Could we figure out the url and return data structure for this for the account service? I could really use it for the mock embedded wallet SDK asap.

My quick guesses:

method: GET
url: /v1/???/[blocknumber | chain?]
Guess at the Response:
```
blocknumber: number;
finalized_blocknumber: number; // ??
genesis: "0x..."; // ?
runtime_version: number; // ??
```

wilwade:
That works for me. BigInt would have to be stringified as json doesn't support, but perhaps for block number we are ok?

PR #872, opened 2025-06-26
Feat/blockinfo endpoint

emresurmeli:
# Problem
Tackling #867

# Solution
This draft implements an endpoint in the blockchain module that allows us to get blockinfo at `/v1/frequency/blockinfo`

# Further points

1. Should we make this controller a part of the api services? ie account and graph each get their own `blockinfo-v1.controller.ts` rather than a shared one in the `/blokchain` folder
2. Potential naming convention changes:
    - Route: `/v1/frequency/block-info`
    - Controller: `chain-info.controller.ts` (we can add more chain related endpoints exposing a lot of the RPC calls through Gateway)
    - Potentially make this its own API called frequency-api which is geared toward exposing chain info. Because this controller being a part of the account-api and graph-api isn't very RESTful?

### DIFF
--- a/apps/account-api/src/api.module.ts
+++ b/apps/account-api/src/api.module.ts
@@ -4,6 +4,7 @@ import { EventEmitterModule } from '@nestjs/event-emitter';
 import { ScheduleModule } from '@nestjs/schedule';
 import { LoggerModule } from 'nestjs-pino';
 import { BlockchainModule } from '#blockchain/blockchain.module';
+import { BlockInfoController } from '#blockchain/blockinfo.controller';
 import { EnqueueService } from '#account-lib/services/enqueue-request.service';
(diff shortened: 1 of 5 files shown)


claireclark1 (approved):
Everything looks good to me! Pattern matching... all of our other controllers are in /apps but this is in /lib. Confirming that's all good.

emresurmeli:
> Everything looks good to me! Pattern matching... all of our other controllers are in /apps but this is in /lib. Confirming that's all good.

This also throws me off - but we can handle that in another issue. I think that it's a bit confusing to have a controller with an endpoint in a part of the repo where it might not belong. But for the sake of putting something out we might want to just go forward with this and allow the SDK to use it quickly.

shannonwells (approved):
Clean, looks good!  I reviewed the code, and I ran gateway and used curl to get the new endpoints for both account-api and graph-api.

Issue #873, opened 2025-06-26
Content Publisher service page returns 404


demisx:
### Where is the Issue?

1. Go to https://projectlibertylabs.github.io/gateway/Build/index.html
2. Click on "Content Publisher" button
3. The https://projectlibertylabs.github.io/gateway/Build/ContentPublishing.html page returns 404. See attached screen below.

<img width="878" alt="Image" src="https://github.com/user-attachments/assets/196e181d-99b7-4523-b0e5-93705571c73f" />

PR #875, opened 2025-06-26
Bug fix the content publishing button href value

emresurmeli:
# Problem 
Issue #873 

# Solution
Ensure we use the right path for the href value on the button.

The false value was: https://projectlibertylabs.github.io/gateway/Build/ContentPublishing.html 
Correct value is: https://projectlibertylabs.github.io/gateway/Build/ContentPublishing/ContentPublishing.html


shannonwells (approved):
I ran mdbook locally and verified the changes.

PR #876, opened 2025-06-27
support ethereum keys for providers #874

### LINKED ISSUE #874: Support ethereum key paths for providers

shannonwells:
# Problem
The previous integration of siwf v2.2.0 didn't take into account provider use of Ethereum keys in their environment config.

Closes #874

# Solution
Where ever we want to use a keypair or address using a Provider config, we have to check what type it is and correctly convert it to values compatible with Frequency.

## not part of solution
A nicer solution would use whatever is in the `.env.*` file to create MSA 1 and make it a Provider but this will have to wait.

I also did not alter any of the k8 tests or other test setup to use ethereum keys.

## Steps to Verify:
1.  `./scripts/account/restart-local-dev.sh`
2.  `npm run start:account-worker:dev` - you should not see any errors thrown at all.
3. Ctrl-C to kill the worker. Then edit `.env.account` and set `PROVIDER_ACCOUNT_SEED_PHRASE='0x5fb92d6e98884f76de468fa3f6278f8807c48bebc13595d45af5bdc4da702133'`
4. `npm run start:account-worker:dev`.  You should see this error thrown; the key is a usable key, it's just not the one for Provider ID 1.
```
[2025-06-26 17:28:52.870 -0700] ERROR (account-worker.main/3008118): [] Provided account secret maps to public key 5HYRCKHYJN9z5xUtfFkyMj4JUhsAwWyvuU8vKB1FcnYTf9ZQ, which does not match configured Provider ID (1)
```

5. Ctrl-C to kill the worker. Then set `PROVIDER_ACCOUNT='deadbeef1234'` or something like that.
6. `npm run start:account-worker:dev`.  You should see this error:
```
[Nest] 3017999  - 06/26/2025, 5:41:00 PM   ERROR [ExceptionHandler] unsupported seed or uri or key type
```


libs/blockchain/src/blockchain-rpc-query.service.ts
```diff
@@ -181,7 +181,7 @@ export class BlockchainRpcQueryService extends PolkadotApiService {
     return {
       blocknumber: await this.getLatestBlockNumber(),
       finalized_blocknumber: await this.getLatestBlockNumber(true),
-      genesis: (await this.api.rpc.chain.getBlockHash(0)).toHex(),
+      genesis: this.api.genesisHash,
```
shannonwells, line 184: I completely overlooked this in my review.  Sorry @hwahby !

shannonwells, line 184: I completely overlooked this in my review, sorry about that.

libs/blockchain/src/capacity-checker.service.ts
```diff
@@ -68,7 +68,7 @@ export class CapacityCheckerService implements OnApplicationBootstrap, OnModuleD
       const capacityInfo = await this.blockchainService.capacityInfo(providerId);
       await this.redis.set(CURRENT_CHAIN_CAPACITY_KEY, JSON.stringify(capacityInfo));
     } catch (err: any) {
-      this.logger.error('Caught error in updateCachedCapacity', err?.stack);
+      this.logger.error(err, 'Caught error in updateCachedCapacity');
```
shannonwells, line 71: For better output. Missed one from the last logger PR

wilwade (approved):
I did a code read through and this looks good from my point of view. It also solves the issue of the derivation path as the private key takes that into account.

Issue #734, opened 2025-03-06
Improve Health Monitoring for Gateway Deployments


mattheworris:
## Description

As a system administrator, I want my deployment to be able to know if a container/pod is up and running. My deployment system should be able to tell if there are issues and restart any containers/pods that are not behaving as expected.

## Acceptance Criteria

- [x] /healthz, /readyz, and /livez should assess the application's functionality as fully as possible
- [x] Either heathz or some other debugging helpful endpoint that answers common questions:
    - mainnet or testnet or other
    - non-sensitive env vars/config values
    - queue statuses
    - utc timestamp
    - current block head or other chain info
    - etc...
- [ ] Both API and Worker applications MUST implement some kind of health check, preferably the above mentioned endpoints.  #896 
- [ ] Helm chart in the docs section is updated to incorporate these checks #897

PR #880, opened 2025-07-01
Expand healthcheck

hwahby:
# Problem

heathz or some other debugging helpful endpoint that answers common questions:

mainnet or testnet or other
non-sensitive env vars/config values
queue statuses
utc timestamp
current block head or other chain info
etc...
problem statement, including
Link to GitHub Issue(s): https://github.com/ProjectLibertyLabs/gateway/issues/734

# Solution

Expand queue status check into a general Redis health check

## Steps to Verify:

1. Run ./start.sh, accept defaults.
2. Make request to verify that the content-publishing-api and content-watcher /healthz endpoints work

- localhost:3010/healthz 
- localhost:3011/healthz 

## Screenshots (optional):

Screenshot 2025-06-30 at 8 01 34 PM


shannonwells:
Can you please be sure to add the new health check endpoints (and also prometheus metrics) to the `🔗💠📡  The selected services are running.` output?  Just occurred to me.  Thanks

apps/account-api/src/api.config.ts
```diff
@@ -1,18 +1,10 @@
...
-  apiTimeoutMs: number;
-  siwfNodeRpcUrl: URL;
-  graphEnvironmentType: keyof EnvironmentType;
-  siwfUrl: string;
-  siwfV2Url?: string;
-  siwfV2URIValidation?: string[];
-}
+import { IAccountApiConfig } from '#types/interfaces/account/api-config.interface';
+
+export { IAccountApiConfig };
```
shannonwells, line 7: Why is this importing and then exporting this interface?

shannonwells, line 7: Question: what's the purpose of importing and then exporting this interface?

hwahby: A pattern we use pretty commonly in the codebase is importing the interface and config together.

This was a way to minimizing changing all the imports. Since the `healthCheckModule` is using the config interfaces, the interface was extracted out (e.g. #types/interfaces/account/api-config.interface';). Reexporting it here was a way to pass it along to where it has been imported.

https://github.com/ProjectLibertyLabs/gateway/blob/main/apps/account-api/src/services/accounts.service.ts#L20

Open to any suggestions to improv the encapsulation of the module/clean up the imports.

shannonwells (approved):
Works for me.  I reviewed the code.  I tested all the endpoints (/healthz, /metrics, and /queues for :3010 - :3015)

Issue #864, opened 2025-06-24
Prometheus endpoint for workers


shannonwells:
The workers don't expose any endpoints for incoming requests, and probably shouldn't, just to serve metrics. The solution is probably to run a sidecar process or a separate module inside the container that serves Prometheus metrics. 

Acceptance criteria
- [ ] figure out a good-security-practices way to serve the metrics and implement
- [ ] tests for same
- [ ] verify that this will work with our deployment setup

Related to #477

PR #899, opened 2025-07-22
feat: add metrics and health endpoints to worker services

hwahby:
# Problem

Service health and Prometheus metrics should be extended to worker services.

problem statement, including
Link to GitHub Issue(s): 
https://github.com/ProjectLibertyLabs/gateway/issues/477
[#864](https://github.com/ProjectLibertyLabs/gateway/issues/864)

# Solution

Setup worker modules as Nest.js "hybrid" applications, creating a http server listening on port 3001. Metrics and Health endpoints could then be registered as normal. 

## Steps to Verify:

1. Start services using local containers (./start.sh or docker-compose per developer docs)
2. Confirm that `/metrics` and `/healthz` endpoints are functional for `account-worker`, `content-publishing-worker`, and `graph-worker`.

- `account-worker: port 3023`
- `content-publishing-worker: port 3020`
- `graph-worker: port 3022`

## Screenshots (optional):

Screenshot 2025-07-28 at 4 45 39 PM

Screenshot 2025-07-28 at 9 42 03 AM

Screenshot 2025-07-28 at 9 41 02 AM

Screenshot 2025-07-28 at 9 40 46 AM


apps/account-worker/src/worker.config.ts
```diff
@@ -3,6 +3,9 @@ import { registerAs } from '@nestjs/config';
 import Joi from 'joi';
 
 export interface IAccountWorkerConfig {
+  apiBodyJsonLimit: string;
+  workerApiPort: number;
```
JoeCap08055, line 7: nit: all other services just call this `apiPort`, maybe rename for consistency (presumably in the other workers as well?)

```suggestion
  apiPort: number;
```

package.json
```diff
@@ -51,9 +51,12 @@
...
+    "generate:openapi:account-worker": "dotenvx run -f env-files/account.template.env -- nest start account-worker -- --writeOpenApi",
     "generate:openapi:content-publishing": "dotenvx run -f env-files/content-publishing.template.env -e ENVIRONMENT=dev -- nest start content-publishing-api -- --writeOpenApi",
+    "generate:openapi:content-publishing-worker": "dotenvx run -f env-files/content-publishing.template.env -e ENVIRONMENT=dev -- nest start content-publishing-worker -- --writeOpenApi",
     "generate:openapi:content-watcher": "dotenvx run -f env-files/content-watcher.template.env -- nest start content-watcher -- --writeOpenApi",
     "generate:openapi:graph": "dotenvx run -f env-files/graph.template.env -- nest start graph-api -- --writeOpenApi",
+    "generate:openapi:graph-worker": "dotenvx run -f env-files/graph.template.env -- nest start graph-worker -- --writeOpenApi",
     "generate:openapi": "npm run generate:openapi:account ; npm run generate:openapi:content-publishing ; npm run generate:openapi:content-watcher ; npm run generate:openapi:graph",
```
JoeCap08055, line 60: need to add workers to this
```suggestion
    "generate:openapi": "npm run generate:openapi:account ; npm run generate:openapi:account-worker ; npm run generate:openapi:content-publishing ; npm run generate:openapi:content-publishing-worker ; npm run generate:openapi:content-watcher ; npm run generate:openapi:graph ; npm run generate:openapi:graph-worker",
```

JoeCap08055 (changes requested):
Looks good, just a couple of things:

* No need to use the Hybrid (microservices) model; the worker apps now should look just like the "-api" apps; ie, just a standard NestExpress application. The only reason we previously used the `createMicroservice` call was because we explicitly did NOT want to listen for HTTP requests. Using two different application setup methods between workers & apis to accomplish the same thing now is just unnecessarily confusing.
* I recommend just using `SERVICE_PORT_X` for the worker ports, rather than a separate set of env vars in the docker-compose files.
* Don't forget to update the following files to add the service port to the workers:
    * `deployment/swarm/docker-compose-swarm.yaml`
    * `docker-compose-published.yaml`
    * `start.sh`

JoeCap08055:
Also, once this is released we'll want to update the `docker-compose-*` files and `start.sh` in the Social App Template repo; might be good to create a tracking issue.

emresurmeli (approved):
lgtm!

apps/account-worker/src/worker.config.ts
```diff
@@ -23,8 +23,8 @@ export default registerAs('account-worker', (): IAccountWorkerConfig => {
       label: 'API_BODY_JSON_LIMIT',
       joi: Joi.string().default('1mb'),
     },
-    workerApiPort: {
-      label: 'WORKER_API_PORT',
+    apiPort: {
+      label: 'API_PORT',
```
JoeCap08055, line 27: `apiPort` is great for the config variable name. Might want to consider a different name for the env var, though. Maybe keep `'WORKER_API_PORT'`? The reason: when using environment files to provision, we typically have a common env file across the api & worker. ie, 'account-api' and 'account-worker' are both driven by 'account.env', so we want to be able to distinguish between the two ports there.

Other solutions to consider:
* use `WORKER_API_PORT` if it exists; otherwise use `API_PORT`
* Split all the environment files so that each app has its own env file in all scripting & Docker setup; ie 'account-api.env' and 'account-worker.env'

PR #907, opened 2025-08-04
fix: do not unreserve nonce if got RpcError for 'Priority is too low'

JoeCap08055:
# Problem
Do not un-reserve the nonce for a failed transaction that failed due to a nonce conflict RpcError.

- [x] Unit tests added

### DIFF
--- a/templates/README.template.md
+++ b/templates/README.template.md
@@ -228,11 +228,9 @@ I would like to thank...
 > (OPTIONAL) Add at least 2 questions new developers would ask when they decide to use your project.
 
 - **[Question_1]**
-
   - [Answer_1]
 
 - **[Question_2]**
-
   - [Answer_2]
 
 <p align="right">(<a href="#-table-of-contents">back to top</a>)</p>
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -60,7 +60,7 @@
       "#openapi": ["libs/openapi/src"],
       "#openapi/*": ["libs/openapi/src/*"],
       "#health-check": ["libs/health-check/src"],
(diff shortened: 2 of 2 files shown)


emresurmeli (commented):
This makes sense. 2 questions:

1. Is it possible for createError() function to throw an exception? If so the nonce will never be unreserved?
2. Could add some logging along the lines of `this.logger.debug("Keeping nonce ${nonce} reserved due to NonceConflictError");` ?

JoeCap08055:
> This makes sense. 2 questions:
> 
> 1. Is it possible for createError() function to throw an exception? If so the nonce will never be unreserved?

I think it's impossible for that function to throw; it literally just constructs some empty class objects that inherit from JS `Error`

> 2. Could add some logging along the lines of `this.logger.debug("Keeping nonce ${nonce} reserved due to NonceConflictError");` ?

If we get a `NonceConflictError` it should be caught by the top-level queue processing for that queue, and logged there. There's a 1:1 relationship between (throw NonceConflictError):(skip unreserving nonce), so I think the extra log would be redundant & noisy.

emresurmeli (approved):

libs/blockchain/src/blockchain.service.spec.ts
```diff
@@ -152,7 +160,6 @@ describe('BlockchainService', () => {
   });
 
   describe('getNetworkType', () => {
-    afterEach(() => {});
```
shannonwells, line 155: That's odd. Any idea what this was doing here?

shannonwells (approved):
Ship it!

Issue #909, opened 2025-08-04
[bug] no error logs in log


JoeCap08055:
# Description
Many queue jobs are erroring out in the Freesky (mainnet) Gateway instance. The BullMQ UI shows us the errors thrown by those failures; according to the source code, error logs should be getting written as well.

shannonwells:
Fixed by #913

PR #913, opened 2025-08-06
Chore/logging cleanup

JoeCap08055:
# Description
This PR reorganizes our logging setup to be more in line with recommended practices for using Pino with NestJS.

* Use dependency injection to utilize a single base logger instead of creating a separate Pino logger instance per class
    * Eliminates excessive resource usage & error messages at app startup (#909)
* App 'main.js' and NestJS instantiation now uses the Pino logging (ie, no switchover to Pino after Nest starts up)
* Monkey-patch override console log methods to funnel through the logger
* All non-class instances of loggers now use the Pino logger as well

Hopefully will also address #909


apps/account-api/src/main.ts
```diff
@@ -93,19 +88,19 @@ async function bootstrap() {
     app.useBodyParser('json', { limit: config.apiBodyJsonLimit });
 
     initializeSwaggerUI(app, swaggerDoc);
-    logger.info(`Listening on port ${config.apiPort}`);
-    logger.info(`Log level set to ${getCurrentLogLevel()}`);
+    logger.log(`Listening on port ${config.apiPort}`);
+    logger.log(`Log level set to ${getCurrentLogLevel()}`);
```
JoeCap08055, line 92: Just for `main.js`, we have to use the Nest-compatible logging API

libs/utils/src/common/common.utils.ts
```diff
@@ -66,3 +67,58 @@ export function validateEnvironmentVariables(logger?: any): void {
...
+              }
+            })(),
+      )
+      .join(' ');
+
+    // now call the original, so it still goes to stderr
+    logger[type](message);
+  };
+
+export function setupLoggingOverrides() {
```
JoeCap08055, line 93: Moved the `BigInt` monkey patch here. Also monkey patch `console.log`, etc, because polkadot-js has a number of console logs that bypass the logger.

testlib/mockPinoLogger.ts
JoeCap08055, line 1: Didn't need to mock this in tests; we just use the real one, it's simple enough.

testlib/redis-provider.mock.spec.ts
JoeCap08055, line 1: There are a few test files that mock Redis on their own, should probably be updated to use this common one.

libs/cache/src/cache-monitor.service.ts
```diff
@@ -3,14 +3,13 @@ import { EventEmitter2 } from '@nestjs/event-emitter';
...
-import { Logger, pino } from 'pino';
-import { getBasicPinoOptions } from '#logger-lib';
+import { PinoLogger } from 'nestjs-pino';
 
 const REDIS_TIMEOUT_MS = 30_000;
 
 @Injectable()
 export class CacheMonitorService implements OnModuleDestroy {
-  private logger: Logger;
+  // private logger: PinoLogger;
```
shannonwells, line 12: nit: remove commented-out code

apps/content-publishing-worker/src/request_processor/dsnp.announcement.processor.spec.ts
```diff
@@ -7,50 +7,54 @@ import { IRequestJob } from '#types/interfaces/content-publishing';
...
+import { LoggerModule } from 'nestjs-pino';
+import { getPinoHttpOptions } from '#logger-lib';
 
 const mockQueue = {
   add: jest.fn(),
 };
 
 // Mock config values
-const mockIpfsConfig: IIpfsConfig = {
+const mockIpfsConfigProvider = GenerateMockConfigProvider<IIpfsConfig>(ipfsConfig.KEY, {
```
shannonwells, line 19: This is a lot of extra change that isn't really related to this PR, which has 72 files.

shannonwells (approved):
I checked out the branch, ran start.sh and verified with swagger api that errors appear in the logs and that all logs are showing up in all services the way I would expect.

libs/cache/src/cache-monitor.service.ts
JoeCap08055, line 12:
```suggestion
```

Issue #915, opened 2025-08-07
Determine how to handle various transaction errors in worker queues


JoeCap08055:
# Description
We need to determine how to handle various error scenarios when processing chain transaction submissions in a worker queue. Failing a worker queue job subjects it to the queues limit on retries (if configured), while sending the job to the "Delayed" queue does not.

Currently, we detect the following conditions when submitting a transaction to the chain (note, once a transaction has been accepted into the pool, it is considered "successful" from the point of view of the worker queue that submitted it, regardless of whether the extrinsic eventually succeeds, fails, or is finalized or not).
* nonce conflict (ie, duplicate nonce)&mdash;in this case, we would expect that re-submission will eventually succeed
* insufficient capacity: here, we typically emit an event that pauses the queue until capacity is refilled or more is staked, so we currently "burn" one failure...
* bad signature: various things can cause this, therefore we have no idea if resubmission would resolve it
* mortality: for some reason the delay between signing & submitting the transaction has been too long (or at least, the block number fetched for the signing was old). Resubmission would likely resolve this... maybe?
* Other RpcErrors... cause/resolution undetermined
* Network error; chain node disconnected ([separate issue](#912); should result in pausing the queue but currently does not)

The question is: is it ever appropriate to utilize the "delayed" queue to resubmit jobs? Currently we do this for duplicate nonces, because when transaction volume is high, this seems to occur more frequently, and until we have a better way to resolve it we don't know for sure how many retries it might take. But, using the delayed queue is currently unbounded; given the same error scenario, the transaction may keep getting retried forever.

JoeCap08055:
## Update
Off-line conversations with @shannonwells and @wilwade determined the path forward:
* Use of the 'delayed' queue is unnecessary
* short-term: all failed jobs follow configured max retry logic for the queue
* longer-term: issue created (ref?) to track potentially making max retries dynamic per job, and environmentally configured per-queue

PR #916, opened 2025-08-07
fix: do not return a Promise inside try...catch

JoeCap08055:
# Problem
A missing `await` was causing errors to be mis-handled.

# Details
Normally, in an `async` Javascript method, it is not necessary to 'await' a value that is being returned; the logic being that since an async function wraps its return value in a Promise anyway, returning a Promise directly is transparent to the caller. So, the following is perfectly fine:
```javascript
async function foo(): Promise<number> {
  // do some async stuff
  return value;
}

async function bar(): Promise<number> {
  return foo();
}
```

In the above, 'bar' doesn't need to 'await foo()', because it will return `Promise<number>` regardless.

However, if we modify the code as follows:
```javascript
async function foo(): Promise<number> {
  // do some async stuff
  if (has_error) {
    throw new Error('I screwed up');
  }

  return value;
}

async bar(): Promise<number> {
   try {
    return foo();
  } catch (err: any) {
    throw Error('foo screwed up');
  }
}
```

In the above, the statement `return foo();` will **always** return a `Promise<number>`. If 'foo' throws, any callers of 'bar', if they 'await' or '.catch(...)' the return value, will only ever get the error 'I screwed up'. The error 'foo screwed up' will **never** be thrown, because 'foo' is always a Promise that hasn't resolved or rejected yet at the time 'bar' returns.

So, the moral of the story is: it's safe to return an encapsulated async function's value directly inside another async function _unless the call is inside a try...catch block_!


JoeCap08055:
## Update
After discussion in issue #915, we eliminated use of the 'delayed' queue, which in turn eliminated the need for the `try..catch`, and thus for the `await`.

PR #920, opened 2025-08-09
feat: add dependency check to readyz request

### LINKED ISSUE #897: Helm Charts add livez/readyz
Split off from #734

The helm charts need to incorporate the use of the livez and readyz endpoints

Related: #896

hwahby:
Closes #897 

# Solution

Expand `/readyz` check to include dependency check. In cases where a dependency was not available, the request would hang. To limit the time the request could hang, I created a new helper util `withTimeoutMs` to attached a timeout to promises. 

Readiness request length is limited to 2 seconds.

Set up common config of default liveness and readiness probe values for k8 helm file. Values set to match existing configurations.

## Steps to Verify:

1. Start services
2. Make request to `/readyz` endpoint
3. Stop Redis or Frequency container
4. Make request to `/readyz` endpoint and confirm "Service is not ready" message

## Screenshots (optional):

Screenshot 2025-08-13 at 11 05 17 AM

Screenshot 2025-08-13 at 11 05 05 AM

### DIFF
--- a/.spellcheckerdict.txt
+++ b/.spellcheckerdict.txt
@@ -133,7 +133,7 @@ Supertest
 svg-embed
 TCP
 Terraform
-(Testnet|TestNet)
+(Testnet|TestNet|testnet)
 TestnetPaseo
 TLS
 Traefik


shannonwells (approved):
This works - briefly - until the other reconnection issue kicks in.  Turning off Redis is the best way to show it but most of the services end up shutting down anyway.  Not your fault of course, it's just something I want to note here.

PR #927, opened 2025-08-15
Get blockTimeMs from BlockchainService to use as the batchAnnoucement batch process window time value constant 

emresurmeli:
Code review feedback implementing a constant that was supposed to be in the PR that introduced the `v3/content/batchAnnoucement` endpoint.

### DIFF
--- a/apps/content-publishing-worker/src/publisher/message.publisher.ts
+++ b/apps/content-publishing-worker/src/publisher/message.publisher.ts
@@ -1,7 +1,6 @@
 import { Injectable, OnApplicationBootstrap } from '@nestjs/common';
 import { SubmittableExtrinsic } from '@polkadot/api-base/types';
 import { DelayedError } from 'bullmq';
-import { BlockchainRpcQueryService } from '#blockchain/blockchain-rpc-query.service';
 import { BlockchainService } from '#blockchain/blockchain.service';
 import { NonceConflictError } from '#blockchain/types';
 import { IPublisherJob, isIpfsJob } from '#types/interfaces/content-publishing';
@@ -24,7 +23,6 @@ export class MessagePublisher implements OnApplicationBootstrap {
   private batchTimeout: NodeJS.Timeout | null = null;
 
   constructor(
-    private blockchainRpcService: BlockchainRpcQueryService,
     private blockchainService: BlockchainService,
     private readonly logger: PinoLogger,
   ) {
@@ -32,11 +30,9 @@ export class MessagePublisher implements OnApplicationBootstrap {
   }
 
   async onApplicationBootstrap() {
-    // Get the real batch size when the application starts
-    this.maxBatchSize = await this.blockchainRpcService.maximumCapacityBatchLength();
-
-    // Wait for blockchain service to be ready to ensure we get actual chain block time
+    // Await blockchain service to be ready before setting batch size and window ms
     await this.blockchainService.isReady();
+    this.maxBatchSize = await this.blockchainService.maximumCapacityBatchLength();
     this.batchProcessWindowMs = this.blockchainService.blockTimeMs;
   }
 


apps/content-publishing-worker/src/publisher/message.publisher.ts
```diff
@@ -9,6 +9,8 @@ import { PinoLogger } from 'nestjs-pino';
 
 @Injectable()
 export class MessagePublisher implements OnApplicationBootstrap {
+  private static readonly BATCH_PROCESS_WINDOW_MS = 6000;
+
```
JoeCap08055, line 13: If the intent is to have the batch window line up with our expected block time, we have a way to get that value from the PolkadotApiServer: `blockTimeMs`

shannonwells: The intent was to start with this but be able to adjust the value if needed.

shannonwells (approved):

apps/content-publishing-worker/src/publisher/message.publisher.ts
emresurmeli: Ah yeah that's better, fixed

JoeCap08055: I wonder if specifying it as a multiplier of the `blockTimeMs` might be a good way to go? I feel like, regardless of the magnitude of the value, the point of using a time interval to collect batches is that it is related to the fact that blocks are also produced on a period, so it's nice to show that relationship in the definition of the constant.

Not important; non-blocking.

apps/content-publishing-worker/src/publisher/message.publisher.ts
```diff
@@ -32,6 +34,7 @@ export class MessagePublisher implements OnApplicationBootstrap {
   async onApplicationBootstrap() {
     // Get the real batch size when the application starts
     this.maxBatchSize = await this.blockchainRpcService.maximumCapacityBatchLength();
+    this.batchProcessWindowMs = this.blockchainService.blockTimeMs;
```
JoeCap08055, line 37: Just need to be careful doing this in the bootstrap code; might need to await the api--the api consts are fetched from the chain when it first connects (though there is a hard-coded fallback value of 6000ms in our code)

emresurmeli: Ah interesting - I noticed that the blockTimeMs is synchronous, so I didn't await it. But I am seeing that there is a isReady async function that checks to see if the api is ready, I can await that and then grab the blockTimeMs.

apps/content-publishing-worker/src/publisher/message.publisher.ts
```diff
@@ -34,6 +34,9 @@ export class MessagePublisher implements OnApplicationBootstrap {
   async onApplicationBootstrap() {
     // Get the real batch size when the application starts
     this.maxBatchSize = await this.blockchainRpcService.maximumCapacityBatchLength();
+
+    // Wait for blockchain service to be ready to ensure we get actual chain block time
+    await this.blockchainService.isReady();
```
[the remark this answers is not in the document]
emresurmeli: I thought about that, but blockchainService is different from blockchainRpcService, so the await won't impact the maximumCapacityBatchLength, which is already being awaited. 

The rpc service also has its own isReady, if we want we could use a similar pattern, but it would be redundant since the maximumCapacityBatchLength function is asynchronous already.

apps/content-publishing-worker/src/publisher/message.publisher.ts
shannonwells: Since the block time is almost never going to change, why get it from the chain?  It's simpler just to set a constant, and then you don't have to worry about any of the above.

JoeCap08055: I'm just concerned that we've got a proliferation of "magic" values that, at the end of the day, are predicated on 6-second block times, so why not make that relationship more obvious?

We actually used to have a constant `BLOCK_TIME_MS` or something like that (env var), but once I realized I could get that value from the chain I eliminated it.

PR #928, opened 2025-08-15
FIX: workers don't shut down gracefully when disconnected from chain

shannonwells:
## Problems
* workers tend to fall over when the chain connection drops, rather than making a graceful exit.
* the BlockchainScannerService is duplicated for each worker, except account services: comparing each revealed minor differences which did not change the behavior.

## Solutions
* always use either a try-catch or a test for connection before querying the chain
* unify BlockchainScannerService where possible.


apps/account-worker/src/transaction_notifier/notifier.service.ts
```diff
@@ -22,7 +22,7 @@ import { PinoLogger } from 'nestjs-pino';
 
 @Injectable()
 export class TxnNotifierService
-  extends BlockchainScannerService
+  extends AccountBlockchainScannerService
```
shannonwells, line 25: Renamed for clarity - this one is unique and uses `BlockchainService`, and _that_ inherits from libs/blockchain/blockchain-scanner.service.  `BlockchainService` does extra things unique to `AccountBlockchainScannerService`.

libs/account-lib/src/utils/account-blockchain-scanner.service.ts
```diff
@@ -37,17 +37,22 @@ export abstract class BlockchainScannerService {
 
   constructor(
     protected cacheManager: Redis,
-    protected readonly blockchainService: BlockchainService,
+    protected readonly blockchainService: BlockchainService, // <<--- NOTE! Not BlockchainScannerService!
```
shannonwells, line 40: I found this easy to miss, hence the comment.

emresurmeli (commented):
Tested account and content publishing workers: 

1. "Chain connected. Unpausing blockchain-scanner service."  logs on workers
2. Once frequency node connection is lost I see graceful disconnection "Chain disconnected. Pausing blockchain-scanner service." in the logs 
3. Once frequency node connection is lost I also see workers are skipping scans instead of crashing "Disconnected: skipping scan"
4. Seeing shutdown timer "starting 60-second shutdown timer"

I notice a couple errors persisting on the workers when the frequency node comes back up

account worker
```
[2025-08-20 21:51:37.866 +0000] ERROR (8072): [RedisModule] NonceService: Connection is closed. 
    err: {
      "type": "Error",
      "message": "NonceService: Connection is closed.",
      "stack":
          Error: Connection is closed.
              at close (/app/node_modules/ioredis/built/redis/event_handler.js:189:25)
              at Socket.<anonymous> (/app/node_modules/ioredis/built/redis/event_handler.js:156:20)
              at Object.onceWrapper (node:events:633:26)
              at Socket.emit (node:events:518:28)
              at TCP.<anonymous> (node:net:346:12)
              at TCP.callbackTrampoline (node:internal/async_hooks:130:17)
```
content publishing worker

```
[2025-08-20 21:51:37.848 +0000] WARN (79): [main] Received shutdown event
[2025-08-20 21:51:37.858 +0000] ERROR (79): [BlockchainRpcQueryService] Failed to reconnect to Frequency nod
e; sending shutdown event.
```

My errors could be specific to how I tested. But reconnection seems to be an issue. Potentially someone could also check it out?

apps/content-publishing-worker/src/monitor/tx.status.monitor.service.ts
```diff
@@ -46,7 +48,7 @@ export class TxStatusMonitoringService extends BlockchainScannerService {
     private readonly capacityService: CapacityCheckerService,
     @InjectQueue(QueueConstants.PUBLISH_QUEUE_NAME) private readonly publishQueue: Queue,
   ) {
-    super(cacheManager, blockchainService, new Logger(TxStatusMonitoringService.prototype.constructor.name));
+    super(cacheManager, blockchainService, new PinoLogger(getPinoHttpOptions()));
```
JoeCap08055, line 51: Why this reversion? Didn't we just clean up all of the `new <some-logger-class>`? We should be injecting a `PinoLogger`, not instantiating a new one.

apps/account-worker/src/transaction_notifier/notifier.service.ts
JoeCap08055: note, `BlockchainService` does not inherit `BlockchainScannerService`

libs/blockchain/src/blockchain-scanner.service.ts
```diff
@@ -70,11 +75,14 @@ export abstract class BlockchainScannerService {
   }
 
   public async scan(): Promise<void> {
+    if (!this.blockchainService.connected) {
+      this.logger.error('Disconnected: skipping scan');
+      return;
+    }
```
JoeCap08055, line 81: See my other comment about this

libs/blockchain/src/blockchain.service.ts
```diff
@@ -89,25 +89,27 @@ export class BlockchainService extends BlockchainRpcQueryService implements OnAp
   }
 
   public async onApplicationShutdown() {
+    this.logger.info('received shutdown event');
```
JoeCap08055, line 92: This log could be misleading; it's not necessarily a foregone conclusion that the app is shutting down due to the `'shutdown'` event. Perhaps:
```suggestion
    this.logger.info('Processing application shutdown');
```

apps/content-publishing-worker/src/monitor/tx.status.monitor.service.ts
shannonwells: Ah right, I just the NestJS Logger, changed it to Pino, forgetting about your recent update.  Will fix.

libs/blockchain/src/blockchain.service.ts
```diff
@@ -89,25 +89,27 @@ export class BlockchainService extends BlockchainRpcQueryService implements OnAp
...
     clearInterval(this.headerInterval);
-    super.onApplicationShutdown();
```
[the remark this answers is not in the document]
shannonwells: TY, it was left over from some older debugging/testing reconnection and I had meant to put it back

apps/account-worker/src/main.ts
```diff
@@ -68,7 +68,7 @@ async function bootstrap() {
   // Get event emitter & register a shutdown listener
   const eventEmitter = app.get<EventEmitter2>(EventEmitter2);
   eventEmitter.on('shutdown', async () => {
-    logger.warn('Received shutdown event');
+    logger.warn('Main Received shutdown event');
```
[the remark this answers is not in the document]
shannonwells: Well, it was helpful for me when I was scanning the logs.

libs/blockchain/src/blockchain-scanner.service.ts
```diff
@@ -118,11 +126,12 @@ export abstract class BlockchainScannerService {
         return;
       }
 
-      // Don't throw if scan paused; that's WHY it's paused
-      if (!this.paused) {
-        this.logger.error(JSON.stringify(e));
-        throw e;
+      // Don't throw if scan paused; just log the error.
```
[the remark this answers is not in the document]
shannonwells: It does.

libs/account-lib/src/utils/account-blockchain-scanner.service.ts
```diff
@@ -87,14 +83,18 @@ export abstract class AccountBlockchainScannerService {
...
-      currentBlockHash = await this.blockchainService.getBlockHash(currentBlockNumber);
-
+      // on a disconnect this gets skipped, a scan in progress will be set to false
+      // in the next block of code, and we exit early.
+      if (!this.paused) {
+        const lastSeenBlockNumber = await this.getLastSeenBlockNumber();
+        currentBlockNumber = lastSeenBlockNumber + 1;
+        currentBlockHash = await this.blockchainService.getBlockHash(currentBlockNumber);
+      }
       if (!currentBlockHash.some((byte) => byte !== 0)) {
```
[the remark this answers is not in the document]
shannonwells: It does throw here, but hits the catch and gets ignored because it's paused.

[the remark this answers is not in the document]
shannonwells: Checking for paused was already in the catch block, and I deduced that that was intentional.  I left it alone ither than the refactor that you and I paired on to make the logic cleaner. If you think it should not be that way, that's fine, but since it's not a true regression, I prefer to address it elsewhere rather tan holding up this PR for another round trip of changes and re-review.  The logs show why the throw occurs. I think we're letting the best be the enemy of the good here.

PR #931, opened 2025-08-26
[WIP] feat: initial ipfs cluster service

### LINKED ISSUE #885: Add support for IPFS Cluster
## Details

While working on Freesky Gateway we went ahead and deployed IPFS cluster just to realize that Gateway currently supports single node mode only. I’ve set `IPFS_ENDPOINT` to the cluster REST API endpoint but it’s sending paths that are IPFS node specific (e.g. `/api/v0/..` in the path):

What’s happening right now:

```sh
# Service is making IPFS Node API calls:
POST http://freesky-ipfs-cluster-local:9094/api/v0/version
POST http://freesky-ipfs-cluster-local:9094/api/v0/pin/add?arg=QmXXXXX
POST http://freesky-ipfs-cluster-local:9094/api/v0/pin/ls
GET http://freesky-ipfs-cluster-local:9094/api/v0/cat?arg=QmXXXXX
```

What we need:

```sh
# Correct Cluster REST API calls:
GET http://freesky-ipfs-cluster-local:9094/version
POST http://freesky-ipfs-cluster-local:9094/pins/QmXXXXX
GET http://freesky-ipfs-cluster-local:9094/pins
POST http://freesky-ipfs-cluster-local:9094/add (for content upload)
I [am confirming with dev team](https://unfinished-team.slack.com/archives/C05M488BL0L/p1752098864521749) whether the Gateway Services only support "single IPFS node" mode right now. If this is the case, we'd need to point to a single IFPS node inside the IPFS cluster as a workaround until the cluster mode is supported in Gateway. I'll update as I hear back anything.
```

Would be great if Gateway supported IPFS Cluster mode as well. Especially important for production scenarios.

hwahby:
# Problem
Add support for IPFS cluster 

Closes #885 

# Solution

### DIFF
--- a/libs/storage/src/ipfs/ipfs.cluster.service.ts
+++ b/libs/storage/src/ipfs/ipfs.cluster.service.ts
@@ -1,0 +1,183 @@
+import { Inject, Injectable } from '@nestjs/common';
+import ipfsConfig, { IIpfsConfig } from './ipfs.config';
+import httpCommonConfig, { IHttpCommonConfig } from '#config/http-common.config';
+import { PinoLogger } from 'nestjs-pino';
+
+@Injectable()
+export class IpfsClusterService {
+  private readonly gatewayUrl: string;
+
+  constructor(
+    @Inject(ipfsConfig.KEY) private readonly config: IIpfsConfig,
+    @Inject(httpCommonConfig.KEY) private readonly httpConfig: IHttpCommonConfig,
+    private readonly logger: PinoLogger,
+  ) {
+    this.logger.setContext(this.constructor.name);
+    this.gatewayUrl = this.config.ipfsGatewayUrl; // Use existing gateway URL
+  }
+
+  public async getVersion(): Promise<any> {
+    const url = `${this.config.ipfsEndpoint}/version`;
+    return this.makeRequest(url, { method: 'GET' });
+  }
+
+  public async addFile(content: Buffer, filename: string): Promise<any> {
+    const url = `${this.config.ipfsEndpoint}/add?pin=true&stream-channels=false`;
+    const formData = IpfsClusterService.createFormData(content, filename);
+    return this.makeRequest(url, {
+      method: 'POST',
+      body: formData.body,
+      headers: formData.headers,
+    });
+  }
+
+  public async getFile(cid: string): Promise<any> {
+    const url = `${this.config.ipfsEndpoint}/cat?arg=${cid}`;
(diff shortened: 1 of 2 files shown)


libs/storage/IPFS-CLUSTER-DESIGN.md
```diff
@@ -0,0 +1,234 @@
...
+  throw err; // Re-throw other errors
+}
+```
+
+## Migration Strategy
+
+Existing deployments can migrate to cluster mode by:
+1. **Environment Update**: Set `IPFS_MODE=cluster` and cluster endpoint
+2. **Service Restart**: No code changes required - configuration injected at runtime
+3. **Gradual Migration**: Services can run in mixed mode during transition
```
JoeCap08055: In the interest of accounting for stupidity, we might want to make it explicit that the IPFS URL that you're pointing to must be a cluster 😛 (ie, you can't just activate cluster mode in Gateway and reap the benefits)
```
Step 1: set IPFS_MODE=cluster
Step 2: ?
Step 3: PROFIT!
```

libs/storage/src/ipfs/ipfs.cluster.service.ts
```diff
@@ -0,0 +1,544 @@
...
+        }
+
+        return Buffer.concat(chunks);
+      }
+    } catch (error: any) {
+      if (error.response) {
+        this.logger.error(`❌ HTTP error! status: ${error.response.status}`, error.response.data);
+      } else {
+        this.logger.error('❌ Error fetching file:', error.message);
+      }
```
JoeCap08055: Use correct argument-passing for Pino logger:

```suggestion
        this.logger.error(error.response.data, `❌ HTTP error! status: ${error.response.status}`);
      } else {
        this.logger.error('❌ Error fetching file: %s', error.message);
      }
```

libs/storage/src/ipfs/ipfs.cluster.service.ts
```diff
@@ -0,0 +1,544 @@
...
+
+    throw lastError;
+  }
+
+  /**
+   * Delay utility for retry backoff
+   */
+  private delay(ms: number): Promise<void> {
+    return new Promise(resolve => setTimeout(resolve, ms));
+  }
```
JoeCap08055: \<rant>You'd think they would have provided an async `delayMs` function or something in Node.js by now instead of always having to do this garbage\</rant>

nit: this is probably better off as a file-scope function instead of a class private. Or better yet, in our utils library?

PR #954, opened 2025-09-04
[DRAFT] stricter eslint rules

shannonwells:
# Problem
This puts back and adjusts most applicable eslint rules from the old airbnb package. Rather than depending on another package's rules, I went through and turned on `recommended` defaults and then adjusted them based on some of the `eslint-disable` comments. I also replaced typescript-eslint plugin with the more up to date package [@typescript-eslint](https://typescript-eslint.io/)

I looked around for new alternatives to the airbnb package. The one I found that seemed reasonable hardly had any rules in it. There are a couple that claim to target NestJS projects, but they were either updated 1-2 years ago or didn't have enough downloads for me to feel good about installing.

# Solution
## Steps to Verify:
1. `npm run lint` should pass - but this is run in CI.

Nitpicking rules is welcome; some of this seems somewhat arbitrary, but I tried to find a happy medium between code quality vs. not having to fix hundreds (literally) of eslint errors.

### DIFF
--- a/eslint.config.js
+++ b/eslint.config.js
@@ -75,7 +75,6 @@ module.exports = defineConfig([
       'no-param-reassign': 'error',
       'no-promise-executor-return': 'error',
       'no-redeclare': 'error',
-      'no-restricted-syntax': 'error',
       'no-restricted-exports': 'error',
       'no-shadow': 'off',
       'no-underscore-dangle': 'error',
--- a/package-lock.json
+++ b/package-lock.json
@@ -128,6 +128,22 @@
         "typescript-eslint": "^8.42.0"
       }
     },
+    "node_modules/@0no-co/graphql.web": {
+      "version": "1.2.0",
+      "resolved": "https://registry.npmjs.org/@0no-co/graphql.web/-/graphql.web-1.2.0.tgz",
+      "integrity": "sha512-/1iHy9TTr63gE1YcR5idjx8UREz1s0kFhydf3bBLCXyqjhkIc6igAzTOx3zPifCwFR87tsh/4Pa9cNts6d2otw==",
+      "license": "MIT",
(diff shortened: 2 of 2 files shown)


apps/account-api/src/controllers/health.controller.ts
```diff
@@ -23,7 +19,7 @@ export class HealthController {
   }
 
   // Live endpoint
-  // eslint-disable-next-line class-methods-use-this
```
shannonwells, line 26: Some of these rules are disabled in so many places, it seemed pointless to turn the rule on.

apps/content-publishing-api/src/controllers/v1/asset.controller.v1.ts
```diff
@@ -48,8 +48,7 @@ export class AssetControllerV1 {
           errorHttpStatusCode: HttpStatus.UNPROCESSABLE_ENTITY,
         }),
     )
-    files: // eslint-disable-next-line no-undef
```
shannonwells, line 51: the no-undef rule is not recommended for use in TypeScript projects.

apps/account-worker/src/transaction_notifier/notifier.service.ts
```diff
@@ -121,7 +121,6 @@ export class TxnNotifierService
         .filter(({ event }) => this.blockchainService.events.capacity.CapacityWithdrawn.is(event))
         .reduce((sum, { event }) => (event as unknown as any).data.amount.toBigInt() + sum, 0n);
 
-      // eslint-disable-next-line no-restricted-syntax
```
shannonwells, line 124: I think the airbnb package must have defined some non-default restricted syntax. I turned it on with just the default settings.

PR #959, opened 2025-09-05
fix: merge bull queue defaults properly

### LINKED ISSUE #958: [bug] Application-wide default BullMQ queue options not being applied correctly
The file [queue.module.ts](https://github.com/ProjectLibertyLabs/gateway/blob/6acc357aacb0656c1eb5e6ca707c168a600c8c0c/libs/queue/src/queue.module.ts) is meant to merge in application-specific queue default options from [queue.constants.ts](https://github.com/ProjectLibertyLabs/gateway/blob/6acc357aacb0656c1eb5e6ca707c168a600c8c0c/libs/types/src/constants/queue.constants.ts), but it seems to be done incorrectly.

Given the following interface:
```typescript
export interface IQueueModuleOptions {
  config?: Partial<QueueOptions>;
  enableUI?: boolean;
  uiRoute?: string;
  queues: RegisterQueueOptions[];
}
```

The `config` property contains the global queue options for all queues. This contains another property, `defaultJobOptions` which controls the default job options for all jobs in all queues, unless overridden at the individual queue or job level.

The `QueueModule` attempts to merge the `config` with some additional properties as follows:
```typescript
      BullModule.forRootAsync({
        useFactory: (cacheConf: ICacheConfig) => ({
          ...options?.config,
          connection: options?.config?.connection || cacheConf.redisOptions,
          prefix: getPrefix(options, cacheConf),
          defaultJobOptions: {
            removeOnComplete: 100,
            removeOnFail: 50,
            attempts: 3,
            backoff: {
              type: 'exponential',
              delay: 2000,
            },
          },
        }),
        inject: [cacheConfig.KEY],
      }),
```

However, the `defaultJobOptions` specified wlil NOT be merged with any `options.config.defaultJobOptions`, but will, in fact, overwrite it completely. Thus any `defaultJobOptions` specified in the global queue config will not be applied; instead, the hard-coded values will be used. Thus the only way to override the default job options is to override for each individual queue or job.

JoeCap08055:
# Problem
The default queue options for Bull queues are not merged correctly.

# Solution
Merge the properties correctly

Closes #958

### DIFF
--- a/libs/queue/src/queue.module.ts
+++ b/libs/queue/src/queue.module.ts
@@ -35,7 +35,8 @@ export class QueueModule {
             removeOnComplete: 100,
             removeOnFail: 50,
             attempts: 3,
-            backoff: {
+            ...options?.config?.defaultJobOptions,
+            backoff: options?.config?.defaultJobOptions?.backoff || {
               type: 'exponential',
               delay: 2000,
             },
--- a/libs/types/src/constants/queue.constants.ts
+++ b/libs/types/src/constants/queue.constants.ts
@@ -70,9 +70,6 @@ export namespace ContentWatcherQueues {
     config: {
       defaultJobOptions: {
         attempts: 3,
-        backoff: {
-          type: 'exponential',
-        },
         removeOnComplete: true,
         removeOnFail: false,
       },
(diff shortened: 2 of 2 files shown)


libs/queue/src/queue.module.ts
```diff
@@ -35,7 +35,8 @@ export class QueueModule {
             removeOnComplete: 100,
             removeOnFail: 50,
             attempts: 3,
-            backoff: {
+            ...options?.config?.defaultJobOptions,
+            backoff: options?.config?.defaultJobOptions?.backoff || {
```
hwahby, line 39: 👍 looks good

hwahby (approved):

PR #963, opened 2025-09-10
Update issue templates

shannonwells:
## Problem
The templates were almost never being used; majority of the time a blank issue was used.
The PR template could be better

## Solution
added some AC checkboxes, added a Bug report template, added default labels, and shortened template text.  Make the PR template more like the Social App Template and the frequency one.


.github/ISSUE_TEMPLATE/content_bug.md
shannonwells, line 1: I think these must have been deleted when the names changed.

emresurmeli (approved):
Ideas addressed 🚢 🚀

PR #975, opened 2025-09-26
fix: get next nonce was repeating last-known nonce

### LINKED ISSUE #974: Spike: investigate frequent  "priority too low" errors
## Problem
The number of failed transactions seems higher than it should be, due to nonce conflicts.  Need to find out if it can be improved.

### Context
These failures fill up the batch queues with failed frequency transaction submissions.  The error looks like:

```
Error: RpcError: 1014: Priority is too low: (430 vs 430): The transaction has too low priority to replace another transaction already in the pool.
```

It happens more often as the transaction submission rate increases; this problem can also crop up in frequency e2e tests, especially if run in parallel mode.

### [PM/Eng]: Acceptance criteria
- [ ] try to quantify when nonce conflicts start happening - a rate of txn submission for example
- [ ] find out whether there is a way to reduce the failures without a major architecture change

JoeCap08055:
# Purpose

The goal of this PR is to reduce or eliminate errors due to nonce conflicts.

Closes #974 

## Solution
There appear to have been two problems:
1. When computing the list of next potentially available nonces, the current nonce was being incremented starting with an index of 0 (ie, `currentNonce + 0 => currentNonce`) instead of 1.
2. When adding the computed offset to the current nonce, 1 was being subtracted (ie, `return currentNonce + offset - 1`, where `offset >= 1`)

Corrected both of these apparent errors.

### Question:
this seems like it was fundamentally broken; ie, would almost always produce an error. How could it ever have submitted a successful transaction?
### Answer (possibly):
 It would always error the first time it submitted a transaction; however, as long as retry/next transaction was submitted before the nonce keys expired from the cache, the next submission would succeed. We may have simply been ignoring errors in the log, since transactions would, more often than not, eventually succeed.

Nevertheless, e2e or load testing of transactions (or at least running the full app & posting transactions) is recommended before releasing to production.

### Checklist

- [x] Unit tests added
- [ ] Integration/end-to-end tests added
- [ ] Documentation added or updated (where applicable)
- [ ] API endpoints added or changed? Swagger docs regenerated
- [ ] Breaking changes? "breaking changes" label added

### DIFF
--- a/libs/blockchain/src/blockchain.service.ts
+++ b/libs/blockchain/src/blockchain.service.ts
@@ -51,15 +51,6 @@ export type Sr25519Signature = { Sr25519: HexString };
 
 const { NUMBER_OF_NONCE_KEYS_TO_CHECK, NONCE_KEY_EXPIRE_SECONDS, getNonceKey } = NonceConstants;
 
-function getNextPossibleNonceKeys(currentNonce: string | number): string[] {
-  const keys: string[] = [];
-  for (let i = 0; i < NUMBER_OF_NONCE_KEYS_TO_CHECK; i += 1) {
-    const key = Number(currentNonce) + i;
-    keys.push(getNonceKey(`${key}`));
-  }
-  return keys;
-}
-
 @Injectable()
 export class BlockchainService extends BlockchainRpcQueryService implements OnApplicationBootstrap {
   private accountId: string;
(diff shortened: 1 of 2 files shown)


libs/blockchain/src/blockchain.service.ts
```diff
@@ -75,6 +66,15 @@ export class BlockchainService extends BlockchainRpcQueryService implements OnAp
     this.readyReject = reject;
   });
 
+  public static getNextPossibleNonceKeys(currentNonce: string | number): string[] {
+    const keys: string[] = [];
+    for (let i = 1; i <= NUMBER_OF_NONCE_KEYS_TO_CHECK; i += 1) {
```
JoeCap08055, line 71: Note changed loop bound from `[0..max]` to `[1..=max]`

Issue #971, opened 2025-09-24
API container runs out of memory when under pressure


demisx:
We are seeing this pattern lately with the Freesky Gateway API container (Content Publishing Service). The memory usage starts under 1Gi, then gradually grows to the container’s limit of 8Gi causing this container to be OOMKilled and restarted. Then, the pattern repeats (see attached image). The Gateway image: `v1.5.10`.

<img width="1595" height="257" alt="Image" src="https://github.com/user-attachments/assets/6d796703-b016-4a88-bbb8-14b71d23d082" />

Attached is the memory usage graph and these are the last messages in the container log right before it’s killed, if helps:

```
🚀 Skipping global interceptor for this route
🚀 Skipping global interceptor for this route
🚀 Skipping global interceptor for this route
🚀 Skipping global interceptor for this route
🚀 Skipping global interceptor for this route
🚀 Skipping global interceptor for this route
🚀 Skipping global interceptor for this route
🚀 Skipping global interceptor for this route
🚀 Skipping global interceptor for this route
🚀 Skipping global interceptor for this route
🚀 Skipping global interceptor for this route
🚀 Skipping global interceptor for this route
🚀 Skipping global interceptor for this route
🚀 Skipping global interceptor for this route
🚀 Skipping global interceptor for this route
🚀 Skipping global interceptor for this route
🚀 Skipping global interceptor for this route
🚀 Skipping global interceptor for this route
🚀 Skipping global interceptor for this route
🚀 Skipping global interceptor for this route
🚀 Skipping global interceptor for this route
🚀 Skipping global interceptor for this route
🚀 Skipping global interceptor for this route
🚀 Skipping global interceptor for this route
🚀 Skipping global interceptor for this route
🚀 Skipping global interceptor for this route
🚀 Skipping global interceptor for this route
🚀 Skipping global interceptor for this route
🚀 Skipping global interceptor for this route
🚀 Skipping global interceptor for this route
🚀 Skipping global interceptor for this route
🚀 Skipping global interceptor for this route
🚀 Skipping global interceptor for this route
🚀 Skipping global interceptor for this route
🚀 Skipping global interceptor for this route
🚀 Skipping global interceptor for this route
🚀 Skipping global interceptor for this route
🚀 Skipping global interceptor for this route
🚀 Skipping global interceptor for this route
🚀 Skipping global interceptor for this route
🚀 Skipping global interceptor for this route
🚀 Skipping global interceptor for this route
[...]

PR #978, opened 2025-09-26
clinic script, k6 script, logging sanitation

shannonwells:
# Purpose
Part of an investigation into #971  brought about some suggested changes here.

1. This called for a k6 load test against content-publishing's `v2/asset/upload` endpoint.
1. prometheus requests massively pollute logs.
1. finding a memory profiling tool that is a little more diagnostic than prometheus was painful and frustrating, but ultimately got something working.

## Solution
1. Write a k6 load test against v2 asset upload endpoint
1. Fix the logs
1. Install [clinic](https://clinicjs.org/) and add a simple script that will make clinic use the k6 test for profiling.
 
### Other notes
In the clinic script, I spent a stupid amount of time trying to convince bash to properly expand a variable and wrap it in single quotes, so one could specify some other k6 script to pass to the `--on-port` parameter on the command line, but it just wouldn't. One way would be make the entire k6 command + file an npm target, but I don't think it's worth adding yet another one.  I figured we won't be doing a lot of memory profiling, so whoever gets in there and does it can just edit the script.

One concern I have about adding clinic, is that it hasn't been updated in a long time, and this added packages with two "critical" and a few "high" npm audit alerts.  At the same time, clinic is a devDependency.

### Steps to Verify
1. You'll need to do some setup steps first (noted in the clinic script), but you should just be able to run scripts/test/clinic.sh as-is.  It will launch content-publishing-api and run the k6 script, profiling memory. When finished it automatically opens a pretty web page with graphs and analysis.
2. Start a local instance of prometheus. Check the log output and verify that NO Prometheus queries are output to the logs.

### Checklist

- [x] Integration/end-to-end tests added
- [x] Documentation added or updated (sort of? in the script.)

## Additional details / screenshot

image

### DIFF
--- a/apps/content-publishing-api/k6-test/script_v2_asset_upload.k6.js
+++ b/apps/content-publishing-api/k6-test/script_v2_asset_upload.k6.js
@@ -1,6 +1,6 @@
 import http from 'k6/http';
 import { check, group } from 'k6';
-import { mockAsset } from './helpers.js';
+import { createMockFile } from './helpers.js';
 
 export const options = {
   vus: 100,
@@ -18,7 +18,7 @@ export const options = {
 export default function () {
   group('/v2/asset/upload large files', () => {
     const url = 'http://localhost:3000/v2/asset/upload';
-    const data = mockAsset('lg', 'image/jpg');
+    const data = createMockFile('lg', 'image/jpg');
     const request = http.post(url, data, { 'Content-type': 'multipart/form-data', accept: 'application/json' });
     check(request, {
       '': (r) => r.status === 202,
--- a/apps/content-publishing-api/k6-test/script_v2_asset_upload.k6.js
+++ b/apps/content-publishing-api/k6-test/script_v2_asset_upload.k6.js
@@ -1,0 +1,27 @@
+import http from 'k6/http';
+import { check, group } from 'k6';
+import { mockAsset } from './helpers.js';
+
+export const options = {
+  vus: 100,
+  duration: '30s',
+  thresholds: {
+    checks: ['rate>=0.995'],
+    http_req_duration: ['avg<400000', 'p(95)<410000'],
+    http_req_failed: ['rate<0.005'],
+    http_reqs: ['rate>=0.24'],
+  },
+  noConnectionReuse: true,
+};
+
+// uses v2 asset upload endpoint to upload a bunch of "large" jpgs
+export default function () {
+  group('/v2/asset/upload large files', () => {
+    const url = 'http://localhost:3000/v2/asset/upload';
+    const data = mockAsset('lg', 'image/jpg');
+    const request = http.post(url, data, { 'Content-type': 'multipart/form-data', accept: 'application/json' });
+    check(request, {
+      '': (r) => r.status === 202,
+    });
+  });
+}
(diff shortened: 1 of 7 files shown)


apps/content-publishing-api/src/api.service.ts
```diff
@@ -252,7 +252,7 @@ export class ApiService {
     const dataOps = await dataTransaction.exec();
     ApiService.checkTransactionResult(dataOps);
     const queuedJobs = await this.assetQueue.addBulk(jobs);
-    this.logger.debug('Add Assets Job: ', queuedJobs);
```
shannonwells, line 255: this was outputting [Object][object]

libs/logger/src/logLevel-common-config.ts
```diff
@@ -28,10 +28,25 @@ export function getPinoTransport() {
   return undefined;
 }
 
+const customLogLevel = (req, res, err) => {
```
shannonwells, line 31: From here https://github.com/pinojs/pino-http?tab=readme-ov-file#logger-options

apps/content-publishing-api/src/api.service.ts
JoeCap08055: Correct usage patterns for the Pino logger with NestJS:

`log.<method>([consolidated object], [message], [...args]);`

If the first argument is an object, it's interpreted as `consolidated object`, ie, a single object containing all of the variables/properties to be logged. ie:
```typescript
const obj1 = { foo: 'hello' };
const obj2 = { bar: 'world' };
this.logger.debug({ obj1, obj2 }, 'This is a log message');
```

results in:
```console
[2025-09-27 13:42:42.653 -0400] DEBUG (97458): [AccountsService] This is a log message
    obj1: {
      "foo": "hello"
    }
    obj2: {
      "bar": "world"
    }
```

If the first argument is a string, there is no "consolidated object", and the first argument is taken as `message`.

Regardless, any arguments following `message` are intended to be arguments for string interpolation within the message, ie:
```typescript
this.logger.debug('This is a log message; here is obj1: %o and obj2: %o', obj1, obj2)
```

results in:
```console
[2025-09-27 13:49:00.087 -0400] DEBUG (284): [AccountsService] This is a log message; here is obj1: {"foo": "hello"} and obj2: {"bar": "world"}'
```
Format specifiers:
%s: String
%d or %i: Integer
%f: Floating-point number
%o or %O: Object (will be serialized to JSON in the log output)
%%: Literal percent sign


I wonder if AI could audit all of our log statements to make sure they follow the pattern `log.<level>([consolidated object], [message], [...args])` (or better yet, actually make the changes)

shannonwells:
> I think this addresses https://github.com/ProjectLibertyLabs/gateway/issues/971
Yes, thanks, @shannonwells should really read more carefully

shannonwells:
@saraswatpuneet I put my notes on the results into the issue #971

apps/content-publishing-api/k6-test/script_v2_asset_upload.k6.js
```diff
@@ -0,0 +1,27 @@
...
+    http_reqs: ['rate>=0.24'],
+  },
+  noConnectionReuse: true,
+};
+
+// uses v2 asset upload endpoint to upload a bunch of "large" jpgs
+export default function () {
+  group('/v2/asset/upload large files', () => {
+    const url = 'http://localhost:3000/v2/asset/upload';
+    const data = mockAsset('lg', 'image/jpg');
```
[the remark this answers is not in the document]
shannonwells: see above

emresurmeli (approved):
🚀

PR #979, opened 2025-09-30
Add default permissions to github actions workflow documents

emresurmeli:
# Purpose

We want to make sure that GitHub actions have default permissions on top of the YAML files that define them. When we add permissions to the top of the file, all jobs below inherit that permission unless they define any permissions within the job.

#877

## Solution

Add code blow to all workflow YAMLs except `deploy-gh-pages.yaml` which has read and write permissions and uses `GITHUB_TOKEN` to auth the writes
```
permissions:
  contents: read
```

### Change summary

* workflow YAMLs get default permissions

### Steps to Verify

1. Check this PRs Github Actions to make sure that the permissions work

### Checklist

- [x] Add default permissions to workflow YAMLs
- [x] Test the Github Actions on a test PR(this one and maybe another)
- [x] Look into Github Token because we might not be setting that on the envs


shannonwells (approved):
Great.  Following only necessary permissions as outlined here, https://graphite.dev/guides/github-actions-permissions

JoeCap08055 (approved):

Issue #731, opened 2025-03-05
Generic Batching for any Batched Schema


wilwade:
As a user of Gateway, I want to be able to batch any schema that is able to be batched without having to add content aware specific endpoints.

## Example
The schemas on testnet were added for the AT Protocol. While the Content Publisher has a way to send a generic schema, it doesn't support batching. The AT Protocol work should work.

jeetumaker:
@wilwade this ticket is critical for our 1.0 sprint planning.  Can you please assign someone on the blockchain team.  I can closely work with them.  Let me know if you wish to have a quick meeting to define the scope and set expectations.

shannonwells:
This winds up being nearly a complete overhaul of content-publisher and content-watcher, because there are enums for each schema, and some are duplicated in places but they are hard coded to the schema ids.  The issue here is that we want to support non-DSNP schemas, and none of those are hard-coded into Gateway. And also it's brittle the way it is.

currently the way the content is handled; the payload format on the rest endpoint is not tied to format of actual on-chain schema.  it's hard-coded to marshal the incoming data into specific fields & manually build the parquet record.

idea 1:  make a generic argument s.t. it will accept a pre-formatted payload for a schema. content-publisher doesn't need to worry about how to format it for the batch file.

idea 2: enable a plugin architecture for gateway so you could deploy app-specific gateway using specific plugins for your needs, that already know about the schemas.  otherwise could default to idea 1, the generic endpoint.

PR #994, opened 2025-10-17
Generic schema batching

emresurmeli:
# Purpose

We want to be able to support batching generic schemas without having to have content aware endpoints.

# Solution

The proposal outlines two approaches. One is a generic schema registry with dynamic discovery (query chain). The other is a hybrid that supports the legacy method of batching schemas (maintaining endpoints) with fallbacks for generic schemas.

Solves #731

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -30,7 +30,7 @@ items that don't apply can be marked NA or deleted
 - [ ] Unit tests added
 - [ ] Integration/end-to-end tests added
 - [ ] Documentation added or updated (where applicable)
-- [ ] API endpoints added or changed? Swagger docs regenerated
+- [ ] API endpoints added or changed? Added the endpoints in main.ts and regenerated Swagger docs
 - [ ] Breaking changes? "breaking changes" label added
 
 ## Additional details / screenshot
--- a/apps/content-publishing-api/src/main.ts
+++ b/apps/content-publishing-api/src/main.ts
@@ -50,7 +50,7 @@ async function bootstrap() {
         'x-tagGroups',
         [
           { name: 'asset', tags: ['v1/asset', 'v2/asset'] },
-          { name: 'content', tags: ['v1/content', 'v2/content'] },
+          { name: 'content', tags: ['v1/content', 'v2/content', 'v3/content'] },
           { name: 'health', tags: ['health'] },
           { name: 'profile', tags: ['v1/profile'] },
         ],
--- a/apps/content-publishing-api/test/app-v3-content-batchAnnouncement.e2e-spec.ts
+++ b/apps/content-publishing-api/test/app-v3-content-batchAnnouncement.e2e-spec.ts
@@ -1,0 +1,213 @@
+import { NestExpressApplication } from '@nestjs/platform-express';
+import { Test, TestingModule } from '@nestjs/testing';
+import { ApiModule } from '#content-publishing-api/api.module';
+import { ApiService } from '../src/api.service';
+import request from 'supertest';
+import {
+  CONFIGURED_QUEUE_NAMES_PROVIDER,
+  CONFIGURED_QUEUE_PREFIX_PROVIDER,
+  ContentPublishingQueues as QueueConstants,
+} from '#types/constants';
+import apiConfig, { IContentPublishingApiConfig } from '#content-publishing-api/api.config';
+import { ValidationPipe, VersioningType } from '@nestjs/common';
+import { TimeoutInterceptor } from '#utils/interceptors/timeout.interceptor';
+import { Logger } from 'nestjs-pino';
+import { EventEmitter2 } from '@nestjs/event-emitter';
+
+describe('AppController E2E request verification', () => {
+  let app: NestExpressApplication;
+  let module: TestingModule;
+
+  beforeAll(async () => {
+    module = await Test.createTestingModule({
+      imports: [ApiModule],
+      providers: [
+        {
+          provide: CONFIGURED_QUEUE_NAMES_PROVIDER,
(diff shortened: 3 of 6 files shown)


docs/proposals/generic-schema-batching-proposal.md
```diff
@@ -0,0 +1,987 @@
...
+│ if batchable:            │
+│   → Batching Queue       │
+│ else:                    │
+│   → Direct Publish       │
+└──────┬───────────────────┘
+       │
+       ▼
+┌──────────────────────────┐
+│ Queue: batch:schema:100  │
+│ • Group by schema ID     │
```
shannonwells: Do you imagine a timeout for publishing a batch if there are no other messages for a batchable schema?

docs/proposals/generic-schema-batching-proposal.md
```diff
@@ -0,0 +1,987 @@
...
+  → No specific adapter found
+  → GenericSchemaAdapter (priority=0)
+  → Pass-through (no transformation)
+  → Queue for Batching
+```
+
+### Benefits of Solution 2
+
+1. **All Benefits of Solution 1** ✅
+2. **Schema-Specific Optimization**: Adapters can implement custom logic, validation, and batch configs per schema
```
shannonwells: Devil's advocate:
- custom logic like validation and formatting could be applied in advance of hitting the endpoint, and community adapters used that way would not need support from us.

shannonwells (commented):
Thanks for your excellent analysis and writeup. There are a couple of spots that aren't precisely correct in understanding about how schemas work, but it doesn't affect your proposals. Solution 2 sounds in many ways really great, but I would ask, what business case or pain point do you have in mind that would it solve right now, or one we anticipate in the near future?

docs/proposals/generic-schema-batching-proposal.md
```diff
@@ -0,0 +1,987 @@
...
+   - Announcement processors
+   - API controllers
+   - Queue mappings
+
+5. **REST API ≠ On-Chain Format**: The current API accepts DSNP Activity Content format but must manually transform it to Parquet format, limiting generic schema support.
+
+### Requirements
+
+**Must Have**:
+- Support batching of any Parquet-based schema
```
JoeCap08055: Q: Do we/should we need to support other than Parquet?

docs/proposals/generic-schema-batching-proposal.md
```diff
@@ -0,0 +1,987 @@
...
+**Location**: `libs/blockchain/src/schema-registry.service.ts`
+
+**Key Responsibilities**:
+- Fetch schema metadata from blockchain via existing RPC methods (`getSchema`, `getSchemaPayload`)
+- Maintain two-way caching: `schemaId ↔ SchemaDefinition` and `schemaName ↔ schemaId`
+- Determine if a schema is batchable by inspecting payload location and model type
+- Pre-load known DSNP schemas at startup for backward compatibility
+- Provide cache invalidation for testing/admin purposes
+
+**Schema Definition Interface**:
```
JoeCap08055: Note, some of the details here will change with the new `Intents` + `Schemas` implementation

docs/proposals/generic-schema-batching-proposal.md
```diff
@@ -0,0 +1,987 @@
...
+**Location**: `apps/content-publishing-api/src/controllers/v1/content.controller.v1.ts`
+
+**New Endpoint**: `POST /v1/content/:msaId/generic`
+
+**Flow**:
+```
+1. Receive request with schemaId + content
+2. Validate schema exists via SchemaRegistry
+3. Check if schema is batchable
+4. If batchable → queue for batching (with schema-specific queue name)
```
JoeCap08055: We might want to consider a different batch/queuing model. One Bull queue per batching schema is not very scalable.

JoeCap08055 (commented):
Overall, I like this. I think Solution 1 is a lay-up. We can then examine our needs and other priorities to decide when/if to proceed with Solution 2.

docs/proposals/generic-schema-batching-proposal.md
```diff
@@ -0,0 +1,987 @@
...
+**Caching Strategy** (Three-Tier):
+```
+Request → In-Memory Map (instant)
+            ↓ miss
+          Redis Cache (~5ms)
+            ↓ miss
+          Blockchain RPC (~50ms)
+            ↓
+          Cache both layers
+```
```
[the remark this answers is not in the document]
shannonwells: For schemas that are outdated/no longer valid, or just haven't been used in awhile, I think a manual endpoint for removing them is a good idea.

docs/proposals/generic-schema-batching-proposal.md
```diff
@@ -0,0 +1,987 @@
...
+
+Replace hard-coded schema mappings with a dynamic registry that fetches schema metadata from the blockchain at startup and on-demand. This foundation enables any schema (DSNP, AT Protocol, custom) to be used without code changes.
+
+### Core Concept
+
+Instead of maintaining a static map of 20 schemas, the system will:
+1. **Query the blockchain** for schema metadata when needed
+2. **Cache aggressively** using a three-tier approach (memory → Redis → blockchain)
+3. **Auto-detect batchability** based on schema settings
+4. **Support any schema** registered on Frequency
```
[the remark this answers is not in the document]
JoeCap08055: I don't think we need to worry about `SignatureRequired` schemas. This generic schema handling is specifically for content batch publishing; publication to the chain is merely announcement of off-chain content. And since batches can consist of entries for multiple users, a signature wouldn't make sense.

While, technically, this could also be used to post `OnChain` content, we don't currently have any schemas that use this, and I don't anticipate any in the near future. In any event, I think we can safely call this out of scope for this PR.

shannonwells (approved):
I think we should go ahead and merge this since it's just a proposal doc, which can be updated later.  I think this will still be of use.  @JoeCap08055 ?

PR #1016, opened 2026-02-03
fix: we were skipping nonces due to JS/Lua array indexing differences

JoeCap08055:
# Purpose

The goal of this PR is to fix a bug that appears to have been introduced by a previous PR. The source of the bug stems from the fact that it was assumed the Redis function `incrementNonce` was returning an offset value potentially starting at zero, but because Redis functions are written in Lua (which indexes arrays from `1`), it was actually returning a minimum value of `1`.

### Steps to Verify

1. Start up any Gateway service that publishes transactions to the chain and submit something
2. Transaction should appear on-chain (previously, was expiring before being included in a block due to a "nonce hole"

### Checklist

- [x] End-to-end tests added

### DIFF
--- a/libs/blockchain/src/blockchain.service.spec.ts
+++ b/libs/blockchain/src/blockchain.service.spec.ts
@@ -277,21 +277,16 @@ describe('BlockchainService', () => {
   describe('reserveNextNonce', () => {
     it('should not fetch from node if current nonce is in cache', async () => {
       jest.spyOn(nonceRedis, 'get').mockResolvedValueOnce('1');
-      const chainSpy = jest.spyOn(blockchainService, 'getNonce').mockResolvedValueOnce(1);
+      const chainSpy = jest.spyOn(blockchainService, 'getNextNonce').mockResolvedValueOnce(1);
       await blockchainService.reserveNextNonce();
       expect(chainSpy).not.toHaveBeenCalled();
     });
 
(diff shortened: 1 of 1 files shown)


libs/health-check/src/health-check.service.ts
```diff
@@ -128,7 +129,7 @@ export class HealthCheckService {
     // If not found in Redis, query the blockchain directly
     try {
       const latestHeader = await this.blockchainService.getLatestHeader();
-      await this.redis.set('latestHeader', JSON.stringify(latestHeader));
+      await this.redis.setex('latestHeader', this.blockchainService.blockTimeMs / MILLISECONDS_PER_SECOND, JSON.stringify(latestHeader));
```
shannonwells: What version of Redis are we using? redis.setex is deprecated:  https://redis.io/docs/latest/commands/setex/

JoeCap08055: Interesting. We're using Redis 7.0.15 (at least, in our Docker setups; in AWS I believe it's Valkey equivalent to Redis 7. In any event, SETEX has been working so far. SETEX still exists for backwards compatibility, at least until v8 (or later); there are no announced concrete plans to remove it at a specific release number.

I'll file an issue in the backlog to replace all instances of SETEX (and any other deprecated commands)

Issue #1034, opened 2026-04-02
Add webhook notification of chain transaction completion to content-publishing-worker


JoeCap08055:
# Description
`content-publishing-worker` should be able to notify clients of completed chain transactions, similar to `account-service-worker`

PR #1037, opened 2026-04-08
Chore/refactor blockchain scanner services

shannonwells:
# Purpose

The goal of this PR is to do some long-needed refactoring of BlockchainScannerService, in preparation for doing #1034.

`AccountBlockchainScannerService` was identical to `BlockchainScannerService`, and that in turn depends on `BlockchainService`, a wrapper for `BlockchainRpcQueryService`, which was unnecessary since it was used by a the `TxNotifierService`, which doesn't need to post transactions.

This change set triggered Copilot to catch an old bug in the notifier. The notifier's new unit tests also serve as regression tests for the bug.

### Change summary
Unifying `*BlockchainScannerService` and making it use only `BlockchainRpcQueryService` required a couple of things.
- `handleSIWFTxnResult` is called by the notifier.  That means it needed to be pulled into `BlockchainRpcQueryService`.
- `handleSIWFTxnResult` needs to access a ProviderId from the config.  The path of least _code changes_ for dealing with this was to just make PROVIDER_ID/ProviderId part of the read-only "non-provider" config. So I renamed it to reflect this change.
- Fix all the broken tests
- Add unit tests for the handleSiwfTxn and notifier service.  These had no tests at all.
- Add PROVIDER_ID to the various configs, k8s, dockers, env templates, etc.
- I added two test suites related to this change set.

### Checklist

- [x] Unit tests added 
- [x] Documentation added or updated (where applicable)
- [x] Docker and k8 configs updated 
- [x] Breaking changes? "breaking changes" label added
- [ ] N/A Submit freesky-infra PR for gateway reflecting new environment variable - I don't believe this is necessary; the only change would be to content-watcher and looking through the repo, it doesn't appear to be deployed. Correct me if I've missed it.


libs/blockchain/src/blockchain-rpc-query.service.ts
```diff
@@ -903,4 +905,48 @@ export class BlockchainRpcQueryService extends PolkadotApiService {
...
+    if (siwfTxnValues.handle === '') {
+      const handle = await this.getHandleForMsa(siwfTxnValues.msaId);
+      siwfTxnValues.handle = `${handle?.base_handle}.${handle?.suffix}`;
+    }
+    if (siwfTxnValues.address === '') {
+      const keyInfo = await this.getKeysByMsa(siwfTxnValues.msaId);
+      siwfTxnValues.address = keyInfo?.msa_keys[0].toString();
+    }
+    if (siwfTxnValues.newProvider === '') {
+      siwfTxnValues.newProvider = this.providerId.toString();
```
shannonwells, line 947: providerId access requiring the change to the blockchain config.

libs/blockchain/src/blockchain-rpc-query.service.ts
```diff
@@ -903,4 +905,48 @@ export class BlockchainRpcQueryService extends PolkadotApiService {
   public async maximumCapacityBatchLength(): Promise<number> {
     return this.api.consts.frequencyTxPayment.maximumCapacityBatchLength.toNumber();
   }
+
+  /**
+   * Handles the result of a SIWF transaction by extracting relevant values from the transaction events.
+   * @param txResultEvents - The transaction result events to process.
+   * @returns An object containing the extracted SIWF transaction values.
+   */
+  public async handleSIWFTxnResult(txResultEvents: FrameSystemEventRecord[]): Promise<SIWFTxnValues> {
```
shannonwells, line 914: This is the function that was moved out of blockchain.service.ts.  This is called in only one place, by the notifier, but IMO it still belongs at the blockchain service level.

libs/blockchain/src/blockchain.service.ts
```diff
@@ -123,50 +123,6 @@ export class BlockchainService extends BlockchainRpcQueryService implements OnAp
     this.headerInterval = setInterval(() => this.updateLatestBlockHeader(), this.blockTimeMs);
   }
 
-  /**
-   * Handles the result of a SIWF transaction by extracting relevant values from the transaction events.
```
shannonwells, line 127: moved to blockchain-rpc-query.service.ts

apps/account-worker/src/transaction_notifier/notifier.service.ts
```diff
@@ -115,7 +112,10 @@ export class TxnNotifierService
     if (extrinsicIndices.length > 0) {
       const epoch = await this.blockchainService.getCurrentCapacityEpoch();
       const events: FrameSystemEventRecord[] = blockEvents.filter(
-        ({ phase }) => phase.isApplyExtrinsic && extrinsicIndices.some((index) => phase.asApplyExtrinsic.eq(index)),
+        ({ phase }) => {
+          return phase.isApplyExtrinsic &&
+            extrinsicIndices.some((index) => phase.asApplyExtrinsic.eq(index))
```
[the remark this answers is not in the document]
shannonwells: note that the logic just below this correctly handles the extrinsicIndex type.

enddynayn (approved):

